PR #30, opened 2015-05-24
Complete the transition of Swappable from variable concept to functio…

CaseyCarter:
…n concept syntax. Follows the style of `EqualityComparable`. Note that I changed the indentation of the `requires` expression in `Swappable<T, U>` but did not add change marks.


ericniebler:
Thanks

PR #41, opened 2015-06-06
Fix bibliography generation

CaseyCarter:
The bibliography isn't being pulled into DXXXX correctly because the Makefile is still trying to `bibtex N4382` instead of `bibtex DXXXX`.


ericniebler:
Thanks

PR #43, opened 2015-06-08
swap and exchange require Destructible

### LINKED ISSUE #40: Should implicit Destructible requirements be made explicit?
I was about to submit a pull request adding `Destructible<T>` to the requirements of `std::swap(T&,T&)` and `std::exchange(T&, U&&)` in [utility] and [utility.swap], when it occurred to me that every standard library function that may destroy an object implicitly requires `Destructible`. Do we want to make those requirements explicit in STL2?

CaseyCarter:
Fixes #40. (Aside: If we decide to decompose Semiregular into Movable and Copyable, Movable is a fairly tight fit for swap & exchange).

### DIFF
--- a/utilities.tex
+++ b/utilities.tex
@@ -10,7 +10,7 @@
 \begin{codeblock}
   // \ref{utility.swap}, swap:
   template<class T>
-    @\newtxt{requires MoveConstructible<T> \&\& MoveAssignable<T>}@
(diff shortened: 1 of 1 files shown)


asutton:
I think that adding Destructible is the wrong way to address this issue. It
would be better to have Copyable and Movable and require Movable for swap()
and exchange. We'll want those concepts when we get to containers anyway.


Andrew Sutton

ericniebler:
You think Movable should require Destructible?

asutton:
Yes.

Andrew Sutton

Issue #44, opened 2015-06-14
Predicate & Relation concepts


CaseyCarter:
I believe the `Predicate` requirement `Convertible<ResultType<F, Args...>, bool>` should in fact be `Boolean<ResultType<F, Args...>>`. If there's a good reason for that specific requirement that I'm missing, then the requirement in `Relation<R, T, U>`:
```
requires (R r, T a, U b) {
  { r(a, b) } -> Boolean;
  { r(b, a) } -> Boolean;
}
```
should be changed to be consistent.

Actually, that requirement in `Relation` needs to be replaced outright with `Predicate<R, T, U> && Predicate<R, U, T>` to not only ensure consistency with the definition of `Predicate`, but to properly require `R` to be regular when applied to `T` and `U`.

ericniebler:
Fixed

PR #45, opened 2015-06-14
Issue 44

CaseyCarter:
Wording to address both problems mentioned in issue #44.


ericniebler:
I like this. I'll merge it when I get to a real computer.

Issue #10, opened 2015-05-06
Should Semiregular be decomposed, as in N3351 Appendix D.1


ericniebler:
Is EqualityComparable overconstraining?
What about Permutable requiring Semiregular?

asutton:
> Is EqualityComparable overconstraining?

No.

> What about Permutable requiring Semiregular?

It probably could, but I haven't done my homework.

CaseyCarter:
My initial thought was that decomposition into Movable and Copyable was inevitable to specify the containers, but in reality few of the container operations move/copy elements directly. `vector` reallocation is the obvious exception; `basic_string` operations are technically also an exception, but `basic_string` requires its elements to be POD. Container operations that move/copy construct elements do so indirectly via their allocators, and the existence of Movable/Copyable has no bearing on MoveInsertable/CopyInsertable. MoveAssignable/CopyAssignable are sufficient for the various flavors of container assignments.

The containers realistically provide little motivation for decomposing Semiregular.

CaseyCarter:
I've given this some more thought in light of the discussion in #43. I would like to propose a decomposition that differs slightly from N3351. Currently in the TS, we have several "feature" concepts that describe individual properties of object types:
* Constructible
* DefaultConstructible
* MoveConstructible
* CopyConstructible
* Assignable
* MoveAssignable
* CopyAssignable

many of which are unusable in isolation, e.g., the capability to copy construct an object isn't useful if that copy cannot be destroyed. They exist to compose object types in the gaps between the concepts in the "object" hierarchy:
* Destructible
* Semiregular
* Regular

This is flexible, but awkward - not all combinations of "features" are sensible, and we end up with weird concepts like Function:
```
concept bool Function =
  Destructible<T> &&
  CopyConstructible<T> &&
  requires (F f, Args...args) {
    typename ResultType<F, Args...>;
    { &f } -> Same<F*>;
    { f.~F() } noexcept;
    { new F } -> Same<F*>;
    { delete new F };
    { f(forward<Args>(args)...) } -> Same<ResultType<F, Args...> >;
  };
```
This design violates one of the fundamental design ideals of N3351 (§1.3):
> The concepts used should express general ideas in the application domain (hence the name
“concepts”) rather than mere programming language artifacts. ... concepts should represent fully formed abstractions.

[...]

PR #51, opened 2015-06-29
Apply the decomposition design from issue #10.

CaseyCarter:
I think this is complete enough to land on master, modulo any final review comments. I consider the side-effect issue (https://github.com/CaseyCarter/stl2/issues/5) to be separate/independent from the decomposition design; wording will be forthcoming in a future pull request.

Err, fixes issue #10 I suppose ;)


ericniebler:
As you can see, I've been working on some corresponding changes to the fundamental concepts in range-v3. I had to play some games there to make things work out. Take a look and let's iterate until we have the two in agreement.

algorithms.tex
```diff
@@ -1560,10 +1563,10 @@ \section*{25.?? Tag specifiers\hfill[alg.tagspec]}\label{alg.tagspec}
 \pnum
 \requires \tcode{Function} shall meet the requirements of
 \tcode{MoveConstructible}~(\changed{Table~\cxxref{moveconstructible}}{
-\ref{concepts.lib.corelang.moveconstructible}}).
+\ref{concepts.lib.object.moveconstructible}}).
 \enternote \tcode{Function} need not meet the requirements of
 \tcode{CopyConstructible}~(\changed{Table~\cxxref{copyconstructible}}{
-\ref{concepts.lib.corelang.copyconstructible}}).\exitnote
+\ref{concepts.lib.object.copyconstructible}}).\exitnote
```
ericniebler: We decided the 'structables are feature concepts. Do we still want to list them in a section for "object" concepts?

concepts.tex
```diff
@@ -13,17 +13,19 @@
...
-based on properties of types.
+based on properties of types. \newtxt{The purpose of these concepts is to establish
+a foundation for equational reasoning in programs.}
 
 \pnum
-The following subclauses describe core language concepts, foundational concepts,
-function concepts, iterator concepts, and rearrangement concepts
-as summarized in Table~\ref{tab:concepts.lib.summary}.
+The following subclauses describe core language concepts, \oldtxt{foundational concepts,}
+\newtxt{comparison concepts, object concepts,} function concepts, iterator concepts,
```
ericniebler: The iterator concepts were in the Iterators chapter. I haven't finished reading yet, so maybe you moved them.

concepts.tex
```diff
@@ -803,6 +564,321 @@
...
+template <class T>
+concept bool Destructible@\newtxt{() \{}\oldtxt{ =}@
+  @\newtxt{return}@ requires (T t@\newtxt{, T* p}@) {
+    { t.@$\sim$@T() } noexcept;
+    @\newtxt{\{ \&t \} -> Same<T*>;}@
+    @\newtxt{delete p;}@
+    @\newtxt{delete[] p;}@
+  };
+@\newtxt{\}}@
+\end{itemdecl}
```
ericniebler: Open question: since Constructible permits reference types, should Destructible also? Rationale in favor: (1) it's ok to let a local reference pass out of scope, and (2) `is_destructible` returns true for reference types.

concepts.tex
```diff
@@ -803,6 +564,321 @@
...
+    @\newtxt{(Destructible<T>() \&\& requires (Args\&\&...args) \{}@
+      @\newtxt{new T\{forward<Args>(args)...\};}@
+    @\newtxt{\}));}@
+@\newtxt{\}}@
+\end{itemdecl}
+
+\ednote{\newtxt{Requiring \tcode{T\{forward<Args>(args)...\}} to be equality preserving
+implies that construction from equal arguments must result in equal objects, i.e., the
+value of a \tcode{Constructible} object must be totally determined by the constructor
+arguments.}}
```
ericniebler: I personally feel we want to loosen this. Andrew's example of a time type that defaults to now is one example.

decomposition.tex
```diff
@@ -0,0 +1,378 @@
...
+in the library, formed by further decomposition of \tcode{Copyable} and \tcode{Movable}. Although
+the Palo Alto Report warns against successive decomposition:
+\begin{quote}
+We recognize that these concepts could be further decomposed into smaller and smaller units.
+For example, we could easily see factoring out shared ``object requirements'' for the
+\tcode{Copyable}, \tcode{Movable}, and \tcode{Function} concepts. However, each subsequent
+refactoring yields concepts with less coherent meaning (if any).
+\end{quote}
+the proverbial ship has already sailed; the standard library uses these concepts in its design
+now. We simply provide convenient names by which to refer to them, and a coherent hierarchy
```
ericniebler: One could argue that this was a mistake that should be remedied. At least, I am curious to know how disruptive it is to the Standard Library to try to make do only with Movable, Copyable, and Semiregular.

decomposition.tex
```diff
@@ -0,0 +1,378 @@
...
+constructed from (possibly \tcode{const}) lvalues or \tcode{const} rvalues with the same value type
+(\tcode{remove_cv_t<T>}). The array allocation requirements forbid \tcode{T} to have explicit
+copy constructor(s).
+
+\rSec2[decomposition.concepts.movable]{Movable (\ref{concepts.lib.object.movable})}
+\begin{codeblock}
+template <class T>
+concept bool Movable() {
+  return MoveConstructible<T>() &&
+    Assignable<T&, T&&>();
```
ericniebler: `Assignable<T&, remove_cv_t<T>&&>()` ??

decomposition.tex
```diff
@@ -0,0 +1,378 @@
...
+
+\rSec2[decomposition.concepts.copyable]{Copyable (\ref{concepts.lib.object.copyable})}
+\begin{codeblock}
+template <class T>
+concept bool Copyable() {
+  return CopyConstructible<T>() &&
+    Movable<T>() &&
+    Assignable<T&, T&>() &&
+    Assignable<T&, const T&>() &&
+    Assignable<T&, const T&&>();
```
ericniebler: remove_cv_t here too?

decomposition.tex
```diff
@@ -0,0 +1,378 @@
...
+\pnum
+The utility of these object concepts versus the syntactic feature concepts should be clear: each
+includes a complement of required semantic behaviors along with its syntactic requirements resulting
+in fully formed abstractions. Similarly to how the strong type system of \Cpp turns many programming
+errors into type errors, the use of these stronger concepts makes it easier to avoid semantic errors.
+Requirements formed using these concepts are both more concise and more strict than the equivalents
+formed by composition of syntactic feature concepts.
+
+\pnum
+Since ``destruction is prerequisite to destruction,'' implicit requirements for destruction become
```
ericniebler: destruction is prerequisite to *con*struction?

concepts.tex
CaseyCarter: That's from N4382:

> The following subclauses describe core language concepts, foundational concepts, function concepts, iterator concepts, and rearrangement concepts as summarized in Table 1.

but fixed nonetheless in e4cae01.

algorithms.tex
CaseyCarter: I strongly feel that Destructible through Regular should be in their own subsection due to the refinements - they are all closely related. I have no strong feelings about the title of that subsection. I did change it from "Value-Semantic Object Concepts" to "Object Concepts" because the structibles no longer have a strong value-semantic character, but I've decided that labeling them as "feature concepts" as I did in that earlier comment is not precisely correct. They provide semantic consistency guarantees making them stronger than the feature concepts in the Standard, e.g., CopyConstructible guarantees destruction, &t == addressof(t), allocation.

"Object concepts" is vague. I'm completely open to an alternate name for this family of concepts but I have no better ideas.

ericniebler:
Great, thanks for doing this Casey.

concepts.tex
CaseyCarter: I've considered it, and ended up in the position of "wait and see." I hate to weaken the concepts without a motivating need, especially disjunctions of unnamed precursor concepts. If we define a concept C = A || B where A and B are internal/unnamed and someone actually cares about the difference between A and B they have to find a way to discriminate the two. If we instead provide A and B as named concepts that problem doesn't exist, and uses that don't need to discriminate can easily form the disjunction of A and B.

It's analagous to my view of type erasure. Yes, type erasure is useful, but in general it's a good idea to push the erasure out toward the edges of the design since it's hard to unerase types.

concepts.tex
CaseyCarter: I think I'm ready to admit now that you are right about this. That property goes too far in that it enforces a particular style of programming which is antithetical to the multi-paradigm design of C++. I've also just realized that `new T{...}` is inherently not equality preserving and needs an exception. I'll push a patch.

ericniebler: > `new T{...}` is inherently not equality preserving.

Good point.

decomposition.tex
CaseyCarter: If `T` is `const` or `const volatile` it makes no difference: both `Assignable<T&, T&&>` and `Assignable<T&, remove_cv_t<T>&&>` *should* be false. If `T` is just `volatile` then I have to admit I'm not sure if either choice is correct. I may want to assign the result of a function that returns `remove_cv_t<T>` to my `T` object, and/or I may want to assign one such object to another. Maybe the safe choice here is the strongest possible requirement: require both `Assignable<T&, T&&>` and `Assignable<T&, remove_cv_t<T>&&>` until/unless usage proves that is the wrong choice?

decomposition.tex
CaseyCarter: Yip.

decomposition.tex
ericniebler: Pathological, but imagine a type `moveonly_proxy<T>` that has a const-qualified `operator=` like:

```c++
const moveonly_proxy& operator=(moveonly_proxy&& that) const &; // proxied move
```

In this case, we want `Movable<const moveonly_proxy>` to be true. I think. And the copies-are-independent semantics only apply to `Copyable` types, so proxies are OK here. It hurts my head and is pathological, so I don't much care. I only suggested it for consistency.

CaseyCarter: > imagine a type `moveonly_proxy<T>` that has a const-qualified `operator=`

How did I get a `const moveonly_proxy&` to assign in the first place? Am I calling the `move` algorithm on a sequence of `moveonly_proxy` with an output iterator that denotes a sequence of `const moveonly_proxy`?

PR #56, opened 2015-07-03
Language for equality-preserving-by-default expressions.

CaseyCarter:
Simplify some of the prose requirements as made possible by equality-preservation of expressions.
Drive-by: Explicitly convert Boolean expressions in prose to bool.
Drive-by: Use "t" and "u" instead of "a" and "b" for the names of the objects of type "T" and "U" in TotallyOrdered<T, U> to be consistent with the other cross-type concepts.
Drive-by: Clarify that the input object is unchanged in Writable and IndirectlyCopyable even if Out is not Readable.
Drive-by: Replace the ugly broken MutableIterator syntax in RandomAccessIterator with equally ugly but hopefully less broken syntax.


iterators.tex
```diff
@@ -348,7 +348,7 @@
   concept bool MoveWritable@\newtxt{() \{}\oldtxt{ =}@
     @\newtxt{return}@ Semiregular<Out>@\newtxt{()}@ &&
       requires (Out o, T v) {
-        *o = move(v);
+        *o = move(v); @\newtxt{// not equality preserving}@
```
ericniebler: Why not?

iterators.tex
```diff
@@ -951,10 +958,12 @@
...
-        (@\xname{MutableIterator}@<I> ?
-          requires(I i, int n) { i[n] = *i; *i = i[n]; } :
-          requires(I i, int n) { { i[n] } -> const ValueType<I>&; });
-      };
+        @\newtxt{\{ i[n] \} -> const ValueType<I>\&;}@
+      @\newtxt{\} \&\&}@
+      (@\newtxt{!}\xname{MutableIterator}@<I> @\oldtxt{?}\newtxt{||}@
+        requires(I i, @\oldtxt{int}\newtxt{DifferenceType<I>}@ n) { i[n] = *i; *i = i[n]; }@\oldtxt{ :}\newtxt{);}@
+        @\oldtxt{requires(I i, int n) \{ \{ i[n] \} -> const ValueType<I>\&; \})}@
+      @\oldtxt{\};}@
```
ericniebler: Better, thanks.

iterators.tex
```diff
@@ -1010,7 +1019,6 @@
     return mem_fn(t);
   }
   template <class T>
-    @\newtxt{requires !is_member_pointer<decay_t<T>{}>::value}@
```
ericniebler: Are you sure?

iterators.tex
CaseyCarter: As the most conservative position since we don't constrain the value or type of that expression to be in any way useful. This could just as well be `void(*o = move(v))` (or `void(*o = v)` for Writable below).

PR #57, opened 2015-07-04
Language requiring stability of equality preserving expressions.

### LINKED ISSUE #38: Should Semiregular impose copy-indenpendence?
I am creating a separate topic for the issue that was mentioned during discussion about `reference_wrapper<I>` in the topic [https://github.com/ericniebler/stl2/issues/23](algorithm return types should maybe drop pair and tuple in favor of structs with more meaningful field names #23).

The question is whether the `Semiregular` concept requires the copy independence, i.e. guarantee that the modification of the copy will not change it source. This may be rephrased into the question if `SemiRegular` types should have value semantics or they are allowed to expose reference semantics.

The non normative node placed in the current _Range TS_ on `Semiregular`:
> The `Semiregular` concept is modeled by types that have behave similarly to built-in types like
int, except that they may not be comparable with `==`. 
Suggest that the types that meets this concept should provide value like semantics (like `int`).

From the other side requiring copy independence from `Semiregular` would break input/output iterators (as pointed by _CaseyCarter_). However I think that the problem should be solved by removing `Semiregular` concept from `WeaklyImplementable`. I believe that such types should not be required to be `CopyConstructible`, because copying the value of the iterator to save current position is undefined for this type of objects, so no algorithm that accepts  these type of iterator needs to create a copy of iterator. To pass is argument to function/return, move operation would be sufficient,

Furthermore copy constructor of `std::istream_iterator` behaves similarly to `std::auto_ptr`- after copy only one of the objects may be used. I think that similar shift like the one in case of std::unique_ptr could be also applied here, which will clearly suggest that now move-from (copied-from before) iterator is not valid.

In conclusion I am suggesting to add copy independence requirements to `SemiRegular` as semantics requirement. And reduce requirement for  `WeaklyImplementable` to `Moveable` and `DefaultConstrutible`.

CaseyCarter:
Supersedes the requirement that Movable types be non-volatile, and provides copy-independence to Copyable. Fixes #38.


concepts.tex
```diff
@@ -49,8 +47,27 @@
...
+further required to be stable: two evaluations of such an expression with the same
+input objects must have the same results absent any intervening direct modification
+of those input objects. \enternote This requirement allows generic code to reason
+about the current values of objects based on knowledge of the prior values as
+observed via equality preserving expressions. It effectively forbids spontaneous
+changes to an object, changes to an object from another thread of execution, changes
+to an object as side effects of non-modifying actions, and changes to an object as
+side effects of modifying a distinct object if those changes could be observable
+to a library function via an equality preserving expression that is required to be
+valid for that object. \exitnote}
```
ericniebler: Notes are non-normative. We don't have any normative description of "direct modification". I think we need one.

CaseyCarter: Yes, that's next: https://github.com/CaseyCarter/stl2/issues/6.

PR #58, opened 2015-07-08
Modification wording.

CaseyCarter:
Some drive-by editorial fixes.

Reword the "multi-pass guarantee" so that it does not require rvalue iterators to be incrementable.

`Sortable` need not specify `ForwardIterator<I>` since `Permutable<I>` already does.

Declare `is_sized_range` and `is_view` normatively so the number of template parameters and class-key are fixed. Define `view_base` normatively so that implementors can't make it non-empty and/or virtual.


CaseyCarter:
This is a fairly big chunk - I'll make time to bang out any issues so we can get this applied before the teleconference on the 13th and people can put eyes on it.

concepts.tex
```diff
@@ -50,24 +57,74 @@
...
+Expressions appearing in a \techterm{requires-expression} in this document
 are required to be equality preserving, except for those annotated with the comment
 ``not equality preserving.'' An expression so annotated may be equality preserving,
-but it is not required to be so.}
+but is not required to be so.
+
+\pnum
+An expression that may alter the value of one or more of its inputs in a manner
+observable to equality preserving expressions is said to modify those inputs.
+This document uses a notational convention to specify which expressions modify what
```
ericniebler: Do you really mean for this convention to apply to the whole document, or just expressions in requires clauses?

concepts.tex
```diff
@@ -50,24 +57,74 @@
...
+\end{codeblock}
+had been declared as well.
+
+Expression \#3 implicitly requires additional expression variants with semantics
+identical to \tcode{a = c} (including non-modification of the second operand),
+as if the expressions \tcode{a = b} and \tcode{a = move(c)} had been declared.
+Expression \#3 does not implicitly require an expression variant with a
+non-constant rvalue second operand, since expression \#2 already specifies exactly
+such an expression explicitly.
+\exitexample
```
ericniebler: @asutton, what think you?

concepts.tex
CaseyCarter: Only expressions in requires clauses. Fixed in https://github.com/CaseyCarter/stl2/commit/d765e211a6bbcc29707e18b8a23787ce3f7a974e.

concepts.tex
CaseyCarter: There's a bit of rationale in https://github.com/CaseyCarter/stl2/issues/6, for what it's worth.

decomposition.tex
```diff
@@ -137,21 +137,21 @@
     // addressability
     { &t } -> Same<T*>;
```
ericniebler: The type of the address of a `const T` is a `const T*`. This raises an issue with your blanket wording about requirements and the const/r-/l-value convention. Sometimes the cv qualification or value category changes the return type in a way that effects the correctness of the constraint as written. I can't think of a clever way to handle this in general. The cases need to be spammed out.

ericniebler: Oh wait. I don't think you meant to change this here. This is the `Movable` concept from the Palo Alto report. But the larger issue remains.

CaseyCarter: > Oh wait. I don't think you meant to change this here. This is the Movable concept from the Palo Alto report. But the larger issue remains.

Yeah, wrong place but right idea. I'll specify all four variants explicitly in `Destructible` and look for similar occurrences. Hopefully something better will come to us later.

iterators.tex
```diff
@@ -276,7 +276,7 @@
   template <class I>
   concept bool Readable@\newtxt{() \{}\oldtxt{ =}@
     @\newtxt{return}@ Semiregular<I>@\newtxt{()}@ &&
-      requires (I i) {
+      requires (@\newtxt{const }@I i) {
```
ericniebler: Minor discomfort. An `optional` is Readable, and it has both const and non-const `operator*` overloads.

iterators.tex
```diff
@@ -347,7 +347,7 @@
   template <class Out, class T>
   concept bool MoveWritable@\newtxt{() \{}\oldtxt{ =}@
     @\newtxt{return}@ Semiregular<Out>@\newtxt{()}@ &&
-      requires (Out o, T v) {
+      requires (@\newtxt{const }@Out o, T v) {
         *o = move(v); @\newtxt{// not equality preserving}@
```
ericniebler: Same thing here. This makes `optional<T>` no longer model `MoveWritable`.

iterators.tex
CaseyCarter: const/non-const overloads - and the rvalue flavors as well - are non-modifying, so that's not a problem.

iterators.tex
CaseyCarter: *This* is definitely a problem. I had a hunch we'd hit on something that modifies on assignment. I'll back it out here and in Writable..

iterators.tex
ericniebler: I know it's not a problem here. It just raised a red flag. For good reason, considering `MoveWritable`.

CaseyCarter: ...except that those overloads cannot have "equivalent semantics", of course. I'll make this modifying, and introduce a non-modifying `operator*` in `WeakInputIterator`.

iterators.tex
```diff
@@ -588,7 +589,7 @@
   template <class I>
   concept bool WeakIterator@\newtxt{() \{}\oldtxt{ =}@
     @\newtxt{return}@ WeaklyIncrementable<I>@\newtxt{()}@ &&
-      requires(I i) {
+      requires(@\newtxt{const }@I i) {
         { *i } -> auto&&; // pre: i is dereferenceable
```
ericniebler: I know this is unrelated to your changes, but this comment hanging out here in space sticks out now. I'll need to do some digging to see if it's even necessary.

iterators.tex
```diff
@@ -1017,7 +1018,7 @@
   template <class T>
     @\newtxt{requires is_member_pointer<decay_t<T>{}>::value}@
   auto @\xname{as_function}@(T&& t@\oldtxt{, std::true_type}@) {
-    return mem_fn(t);
+    return mem_fn(@\newtxt{forward<T>(}@t@\newtxt{)}@);
```
ericniebler: If `T` is a member pointer, it doesn't need to be forwarded.

iterators.tex
CaseyCarter: There's no nearby normative definition of "dereferencable", IIRC. If we make one, it should be up near Readable et al.

iterators.tex
```diff
@@ -1638,7 +1641,7 @@
 \begin{addedblock}
 \begin{codeblock}
   template <class C> auto size(const C& c) -> decltype(c.size());
-  template <class T, size_t N> constexpr size_t begin(T (&array)[N]) noexcept;
+  template <class T, size_t N> constexpr size_t @\oldtxt{begin}\newtxt{size}@(T (&array)[N]) noexcept;
```
ericniebler: Good catch.

iterators.tex
```diff
@@ -6370,6 +6380,11 @@
 may want to specialize \tcode{is_sized_range} to override the default in the
 case of accidental conformance.\exitnote
 
+\ednote{\newtxt{This language needs to be stronger than ``free to/may want
+to specialize.'' If \tcode{size(foo const\&)} launches the missiles, then a
+user either \emph{must} specialize \tcode{is_sized_range<foo>} or we need
+specify to which library functions that user cannot pass a \tcode{foo}.}}
+
```
ericniebler: Right. If a type does not meet the semantic requirements of a `SizedRange` (namely that `size(r)==distance(begin(r), end(r))`, modulo signedness), then `is_sized_range` must be specialized.

iterators.tex
CaseyCarter: Fair enough, my "forwarding reference" reflexes kicked in.

iterators.tex
```diff
@@ -6447,9 +6465,10 @@
 template <class T>
 struct is_view : false_type { };
 
-template <Range T>
-  requires Derived<remove_reference_t<T>, view_base>@\newtxt{()}@ ||
-    !@\xname{ContainerLike}@<remove_reference_t<T>>
+template <@\oldtxt{Range}\newtxt{class}@ T>
+  requires @\newtxt{Range<T>() \&\&}@
+    @\newtxt{(}@Derived<remove_reference_t<T>, view_base>@\newtxt{()}@ ||
+      !@\xname{ContainerLike}@<remove_reference_t<T>>@\newtxt{)}@
```
ericniebler: Why?

CaseyCarter: I somehow convinced myself this was a concept definition. Will revert.

decomposition.tex
CaseyCarter: Correction: all two variants. I don't think we should to require `&` for rvalues.

iterators.tex
CaseyCarter: ...and now I'm waffling on this again. As far as Readable is concerned, the semantics of `*` are that it requires the value to be dereferencable and that it returns something convertible to `const ValueType<I>&`. If `*declval<T&>()`, `*declval<const T&>()`, `*declval<T&&>()`, and/or `declval<const T&&>()` return different types they can still be "semantically equivalent" if they are all non-modifying, return something convertible to `const ValueType<I>&`, and the values returned are not observably different to equality preserving expressions. In the case of `optional<U>`, the returned types would be `U&`, `const U&`, `U&&` and `const U&&` all bound to the same object. I think we can make `*i` non-modifying with the full complement of expression variants and not lose even Readables like `optional<U>` with deep constness.

ericniebler: Does "semantically equivalent" mean they must not have any observable side-effects that are different? If `*declval<T&>()`, `*declval<const T&>()`, `*declval<T&&>()`, and/or `*declval<const T&&>()` return different types, are all non-modifying, and return something convertible to `const ValueType<I>&`, can those conversion operators write different things to `cout`, or would `T` no longer model `Readable`?

Do we define "semantic equivalence" anywhere? Side-effects are allowed by equality-preserving expressions. They should probably be allowed here.

Perhaps we can tie the definition of "semantic equivalence" to "equality preserving".

CaseyCarter: We use "identical semantics" in the description of "implicitly declared expression variants" and "semantically identical to" in the requirements for Boolean. Without definition. I'm reasonably certain that we mean something like "equivalent as observable by equality-preserving expressions" which is a bit looser than the standard's notion of "observable side effects."

CaseyCarter: > Two expressions are semantically equivalent when they return equal values and have the same effects - if any - on input objects. [Note: Semantically equivalent expressions may have different side effects.] [Examples: `((void)i++, i)` is equivalent to `(++i)` when `i` is `Incrementable`. `bool(a == b)` is equivalent to `bool(a) == bool(b)` when `a` and `b` are the same model of `Boolean`. If `v` is an object of type `std::vector<int>`, `*std::begin(v)` is equivalent to `*v.begin()` and `*std::cbegin(v)` and `v.front()`.]

ericniebler: The examples should use only built-ins. I also wonder if the committee will balk at our defining "semantically equivalent" to allow for observably different side-effects. But similar allowance is made for copy elision under the as-if rule. Hrm, note to self: look into the as-if rule and see if it can be used here.

CaseyCarter: > I also wonder if the committee will balk at our defining "semantically equivalent" to allow for observably different side-effects.

Almost certainly yes. The "as-if" rule is the closest thing to semantic equivalence in the standard, and it covers all observable effects. This weaker definition probably can't be applied usefully to the rest of the standard. I think instead of trying to define "semantically equivalent" to be what I need for "implicit expression variants", I should explicitly characterize the requirements for expression variants. What I really need is substitutability: the expression variants must satisfy the requirements of the primary expression. A couple of sentences to describe that is probably a more feasible goal than trying to generally define "semantically equivalent."

ericniebler:
IIUC, the only outstanding issue to be resolved with this is the wording for semantic equivalence, is that right? That and squashing the commits.

CaseyCarter:
> the only outstanding issue to be resolved with this is the wording for semantic equivalence

Yes. I've been thinking about the wording while I work on getting cmcstl2 to compile preparatory to converging both range-v3 and cmcstl2 with the TS. I think that requiring implicit expression variants to meet the requirements of the declared expression is the right answer. I'll word something now so we can get this closed.

CaseyCarter:
@ericniebler There we go, "semantic equivalence" eradicated so we need not define it.

ericniebler:
OK, good. Squash it, and I'll merge.

CaseyCarter:
Squish. I'll be attending the teleconference on Monday, so you can blame any errors on me ;)

ericniebler:
thanks!

asutton:
Where did this end up? I saw that "semantic equivalence" was removed
(yeah!), but I haven't seen the new wording.

I was about the definition of "semantic equivalence" because I went through
3 months of back and forth when another author of the Palo Alto TR on
similar concepts (behaviorally equivalent, equivalent observable effects,
etc).

None of those ideas work for C++. Throw some debugging code in a function,
and you're suddenly (morally) invoking undefined behavior because your
observable effects differ.

The only thing we can require of expressions is equality preservation.

Andrew

CaseyCarter:
> I was about the definition of "semantic equivalence" because I went through
3 months of back and forth when another author of the Palo Alto TR on
similar concepts (behaviorally equivalent, equivalent observable effects,
etc).

We came to the same conclusion - trying to define "semantically equivalent" to mean "completely and totally identical but only in the ways we actually care about" was a dead end. The conclusion was to explicitly describe the "things we actually care about" in specific instances and avoid a general definition altogether.

asutton:
Good news! Glad it took less than 3 months :)


Andrew Sutton

PR #60, opened 2015-07-15
Split Constructible into __ConstructibleObject and __BindableReference.

CaseyCarter:
T{args...} does not work when T is a reference type ([CWG1521](http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_active.html#1521)). Also, this "decomposed" presentation is more readable.

Closes CaseyCarter/stl2#12.


asutton:
How much of this complexity is driven by the fact that concepts doesn't
allow "declaration constraints". What if you could write:

T t { args... };


Andrew Sutton

CaseyCarter:
> How much of this complexity is driven by the fact that concepts doesn't
allow "declaration constraints"?

Very little, really. `Constructible` is inherently a disjunction of "object type constructible from argument types" and "reference type bindable to argument types." The only reason they shared the expression `T{args...}` in the first place was an attempt on my part to shorten the monster's syntax. I think this new presentation makes it much clearer that `Constructible` is two things in one and what the purposes of those two things are. Even if we could write `T t{args...}` I think I'd leave `Constructible` decomposed as it's more comprehensible.

concepts.tex
```diff
@@ -728,15 +728,26 @@
 
 \indexlibrary{\idxcode{Constructible}}%
 \begin{itemdecl}
+@\newtxt{template <class T, class ...Args>}@
+@\newtxt{concept bool \xname{ConstructibleObject} = // exposition only}@
+  @\newtxt{Destructible<T>() \&\& requires (Args\&\& ...args) \{}@
+    @\newtxt{T\{forward<Args>(args)...\} // not equality preserving}@
```
ericniebler: I'm in the process of manually merging this, but for future reference, we're explicitly qualifying `std::forward` and `std::move` now.

concepts.tex
```diff
@@ -728,15 +728,26 @@
 
 \indexlibrary{\idxcode{Constructible}}%
 \begin{itemdecl}
+@\newtxt{template <class T, class ...Args>}@
+@\newtxt{concept bool \xname{ConstructibleObject} = // exposition only}@
```
ericniebler: Use `\expos` instead of "exposition only".

concepts.tex
```diff
@@ -728,15 +728,26 @@
...
+  @\newtxt{Destructible<T>() \&\& requires (Args\&\& ...args) \{}@
+    @\newtxt{T\{forward<Args>(args)...\} // not equality preserving}@
+    @\newtxt{new T\{forward<Args>(args)...\} // not equality preserving}@
```
ericniebler: This is now spelled, "not required to be equality preserving".

PR #66, opened 2015-07-23
The return value of swap - if any - is not interesting.

CaseyCarter:
As noted in ericniebler/range-v3#204.


concepts.tex
```diff
@@ -326,7 +326,7 @@
 template <class T>
 concept bool Swappable() {
   return requires(T t, T u) {
-    swap(std::forward<T>(t), std::forward<T>(u));
+    @\newtxt{(void)}@swap(std::forward<T>(t), std::forward<T>(u));
```
ericniebler: Do we have a note anywhere explaining the significance of casting to `void` wrt equality preservation?

CaseyCarter: Do you think so? Would changing [concepts.lib.general.equality]/1 from
> The output of an expression is the expression’s result and all operands modified by the expression.
to
> The output of an expression is the expression’s result, if non-`void`, and all operands modified by the expression.
suffice?

Issue #67, opened 2015-07-28
SizedRange and is_sized_range


CaseyCarter:
If I want to use my range type `R` which meets the syntactic requirements of but does not model `SizedRange` with the standard library, am I required to specialize `is_sized_range<R>` to `false_type`? Or am I required to specialize `is_sized_range<U>` to `false_type` for all `U` such that `is_same_v<uncvref_t<U>, R>` is `true`? The current wording would seem to indicate the latter.

I think `is_sized_range` should be insensitive to cv-quals/references for user convenience. That makes implementation of the default complicated since `Range` and hence `SizedRange` are sensitive to cv-quals/references. It's more easily implemented as:
```
namespace detail {
template <class T>
concept bool SizedRangeLike =
  Range<T>() &&
  requires (T& t) {
    { size(t) } -> Integral; // non-modifying
    requires ConvertibleTo<decltype(size(t)), DifferenceType<IteratorType<T>>>();
  };  
}

template <class>
struct is_sized_range : std::true_type {};

template <class T>
concept bool SizedRange() {
  return meta::_v<is_sized_range<uncvref_t<T>>> &&
    detail::SizedRangeLike<T>;
}
```
Thoughts?

ericniebler:
I prefer an implementation like [that in range-v3](https://github.com/ericniebler/range-v3/blob/master/include/range/v3/range_concepts.hpp#L418). `is_sized_range<X>` and `is_sized_range<X const &>` should return the same result always, and they should always default to `false` if `X` is not a range.

```c++
template <class X>
struct __is_sized_range : std::false_type {};

template <class T, std::size_t N>
struct __is_sized_range<T[N]> : std::true_type {};

template <__SizedRangeLike X>
struct __is_sized_range<X> : std::true_type {};

template <class T>
concept bool _Unqualified = Same<T, __uncvref_t<T>>();

template <class T>
struct is_sized_range : is_sized_range<__uncvref_t<T>> {};

template <_Unqualified T>
struct is_sized_range<T> : __is_sized_range<T> {};
```

ericniebler:
Implemented in  6411f9a

CaseyCarter:
[...]

PR #69, opened 2015-07-29
Update wording for SizedRange. Closes issue ericniebler/stl2#67

CaseyCarter:
The specification of `is_sized_range` is a bit awkward, I think it may be better to simply specify the definition normatively.


iterators.tex
```diff
@@ -6373,19 +6373,22 @@
...
 template <class T>
 concept bool @\xname{SizedRangeLike}@ =
-  Range<T>@\newtxt{()}@ &&
-  requires(@\newtxt{const }@T t) {
+  Range<T@\newtxt{\&}@>@\newtxt{()}@ &&
+  requires(@\newtxt{const }@T@\newtxt{\&}@ t) {
     { size(t) } -> Integral;
-    requires Convertible@\newtxt{To}@<decltype(size(t)),
-                           DifferenceType<IteratorType<T>>>@\newtxt{()}@;
+    @\newtxt{\{ size(t) \} -> ConvertibleTo<DifferenceType<IteratorType<T\&>{}>{}>;}@
```
ericniebler: Placeholders FTW!

iterators.tex
```diff
@@ -6394,26 +6397,32 @@
 only if \tcode{size(t)} returns the number of elements in \tcode{t}.}
 
 \pnum
-For any type \tcode{T}, \tcode{is_sized_range<T>} derives from
-\tcode{true_type} if \tcode{T} \oldtxt{models}\newtxt{meets the syntactic requirements of}
-\tcode{\xname{SizedRangeLike}}, and \tcode{false_type} otherwise.
+For a\oldtxt{ny} type \tcode{T} \newtxt{that satisfies \tcode{\xname{SizedRangeLike<T>}}},
+\tcode{is_sized_range<T>} derives from \tcode{true_type} if \oldtxt{\tcode{T} models
+\tcode{\xname{SizedRangeLike}}} \newtxt{\tcode{Same<T, remove\_cv\_t<remove\_reference\_t<T>{}>{}>()} is satisfied}, and \oldtxt{\tcode{false_type}} \newtxt{\tcode{is_sized_range<remove_cv_t<remove_reference_t<T>{}>{}>}} otherwise.
```
ericniebler: We have a problem that the name of an exposition-only concept is appearing in normative text. We might need to promove `_SizedRangeLike` to being a full-fledged part of the design, but I'm not happy about that. I'm open to suggestions.

iterators.tex
```diff
@@ -6394,26 +6397,32 @@
...
 \pnum
 Users are free to specialize \tcode{is_sized_range}.
 \enternote \newtxt{As indicated in the requirements for the \tcode{Range}
 concept above,} users \oldtxt{may want to}\newtxt{must} specialize
 \tcode{is_sized_range} \newtxt{so that its member \tcode{value} is \tcode{false}}
-to override the default in the case of accidental \newtxt{syntactic}
-conformance. \exitnote
+to override the default \oldtxt{in the case of accidental conformance}
+\newtxt{for \tcode{Range} types that satisfy \tcode{\xname{SizedRangeLike}} but
+do not satisfy \tcode{SizedRange}}. \exitnote
```
ericniebler: We could be more specific and say "do not satisfy the extra semantic requirements of `SizedRange` or some such since `_SizedRangeLike` and `SizedRange` differ only in their semantic requirements.

iterators.tex
```diff
@@ -6394,26 +6397,32 @@
...
-struct is_sized_range : false_type { };
+@\newtxt{// \expos}@
+@\newtxt{template <class T>}@
+@\newtxt{using \xname{uncvref} = remove_cv_t<remove_reference_t<T>{}>;}@
+
+template <@\oldtxt{class}\newtxt{\xname{SizedRangeLike}}@ R>
+struct is_sized_range : @\oldtxt{false_type}\newtxt{is_sized_range<\xname{uncvref}<R>{}>}@ { };
 
 template <@\xname{SizedRangeLike}@ R>
+  @\newtxt{requires Same<R, \xname{uncvref}<R>{}>()}@
```
ericniebler: I expect we'll want to use `__uncvref_t` elsewhere (e.g. when defining `is_view` which also needs work). Maybe we could put it in the library introduction?

iterators.tex
CaseyCarter: This should probably be simply `{ size(t) } -> DifferenceType<IteratorType<T&>>`, but this way we won't need to change it after `ConvertibleTo` is fixed to require implicit and explicit conversion with equivalent results.

iterators.tex
CaseyCarter: I suppose the alternative would be to specify the requirements directly in the text instead of naming `SizedRangeLike`, but eww no.

It's part of the design whether the name starts with _ or not. It seems odd to me to fully specify this concept which users critically must understand to avoid undefined behavior and every implementation needs to implement almost exactly as specified yet keep it exposition only. It's there, I know that it exists and I have to conform to it, but I can't rely on the name.

Of course we wouldn't need it at all if we changed to a pure opt-out trait:
```
template <class R>
struct disable_sized_range : std::false_type {};
template <class R>
concept bool SizedRange() {
  return _IsNot<__uncvref<R>, disable_sized_range> && Range<R> && requires (const R& r) {
    { size(r) } -> Integral;
    { size(r) } -> ConvertibleTo<DifferenceType<IteratorType<R&>>>;
  };
}
```
Gosh that sure does make the design simple ;)

ericniebler: > Of course we wouldn't need it at all if we changed to a pure opt-out trait.

It doesn't seem to me that we could handle `is_view` in a way that is consistent with that.

CaseyCarter: Right, since `is_view` is both opt-out and opt-in.
```
template <class T>
struct enable_view {};
template <class T>
concept bool View() {
  return Range<T>() &&
    Semiregular<T>() &&
    (_Is<__uncvref<T>, enable_view> ||
     (!meta::has_type<enable_view<__uncvref<T>>>::value &&
      (DerivedFrom<__uncvref<T>, view_base> || !ContainerLike<T>)));
}
```
Although that clause with the disjunctions is hideous, and probably needs to become a predicate.

iterators.tex
CaseyCarter: Agree.

iterators.tex
ericniebler: I see what you did there. That isn't too awful, I suppose. The bit with `meta::has_type` could also be written `_Valid<meta::_t, enable_view<__uncvref<T>>>`. For the specification, we can't refer to `meta`, though. *&lt;sigh&gt;* C++ needs some metaprogramming facilities.

> Although that clause with the disjunctions is hideous, and probably needs to become a predicate for performance. 

Why not put the derived-from/!container-like logic into the `enable_view` trait?

Another approach I've sometimes used in cases like this is make the primary template inherit from a special type. That way, there's a simple way to detect whether the template has been specialized. Not sure if that helps here.

FWIW, I think I'm now convinced.

CaseyCarter: > For the specification, we can't refer to meta, though.

There's always *see below* ;)

> Why not put the derived-from/!container-like logic into the enable_view trait?

Mostly paranoia about putting logic of any kind in a user-specializable template. It's generally easier to avoid doing so than to determine in what ways a user may bypass that logic and break the design.

CaseyCarter: > FWIW, I think I'm now convinced.
Take a look at https://github.com/CaseyCarter/cmcstl2/compare/SRP_traits and I'll make changes to this wording.

ericniebler: I like it.

ericniebler: >> For the specification, we can't refer to meta, though.
>
> There's always see below ;)

One of these days I'll propose Meta for the standard library. It would make so many things so much easier.

iterators.tex
```diff
@@ -6369,54 +6369,73 @@
...
 
 \pnum
-For any type \tcode{T}, \tcode{is_sized_range<T>} derives from
-\tcode{true_type} if \tcode{T} \oldtxt{models}\newtxt{meets the syntactic requirements of}
-\tcode{\xname{SizedRangeLike}}, and \tcode{false_type} otherwise.
+For any type \tcode{T}, \tcode{\oldtxt{is}\newtxt{disable}_sized_range<T>}
+derives from \oldtxt{\tcode{true_type} if \tcode{T} models
+\tcode{\xname{SizedRangeLike}}, and} \tcode{false_type} \newtxt{if \tcode{T}
+is a non-reference cv-unqualified type, and from \\
+\tcode{disable_sized_range<remove_cv_t<remove_reference_t<T>{}>{}>}} otherwise.
```
CaseyCarter: This may be overkill. I can't imagine a range with `size` valid for some combination of qualifiers but not for all combinations of qualifiers. We could simplify this to `template <class> struct disable_sized_range : false_type {};` and I double anyone would ever notice the difference.

ericniebler: We have to check ``!disabled_size_range<__uncvref_t<T>>::value` in `SizedRange`, right?

CaseyCarter: Yip. The current formulation allows one to disable `SizedRange` for e.g. `const volatile T&` while leaving it enabled for other combinations of qualifiers, or to disable generally for all combinations of qualifiers by specializing `disable_sized_range<T>`, or to disable generally and then re-enable specific combinations by specializing `disable_sized_range<volatile T&&>` to `false_type`.

If we just check `!disable_sized_range<__unvref_t<T>>::value` then it's all-or-nothing.

iterators.tex
```diff
@@ -6442,56 +6461,84 @@
...
-\begin{codeblock}
+\begin{itemdecl}
 @\newtxt{template <class T>}@
-@\newtxt{struct is_view;}@
+@\newtxt{struct enable_view \{ \};}@
 
 @\newtxt{struct view_base \{ \};}@
 
+@\newtxt{template <class T>}@
+@\newtxt{constexpr bool \xname{view_predicate}; // \expos}@
```
ericniebler: Is this valid C++? I think `constexpr` objects need to be initialized. Might need an `= /seebelow;`.

CaseyCarter: [dcl.constexpr]/1:

> The `constexpr` specifier shall be applied only to the definition of a variable or variable template, the
declaration of a function or function template, or the declaration of a static data member of a literal
type (3.9).

So no, it is not valid.

iterators.tex
ericniebler: There's a difference between defining `struct disable_sized_range : disable_sized_range<__uncvref_t<T>>` and testing `!disable_sized_range<__uncvref_t<T>>::value`. With the later, the user can, if she really wants, override the default for particular qualified types. My gut tells me we want to give the users that control, but my gut has been wrong before.

iterators.tex
ericniebler: Also, `__view_predicate` is a strange name. Why not `__is_view`?

iterators.tex
CaseyCarter: > My gut tells me we want to give the users that control, but my gut has been wrong before.

No strong feelings either way. I was thinking about changing the two-step check into a three-step check (`T` => `remove_reference_t<T>` => `uncvref_t<T>`) and it occurred to me that even two steps is likely overkill. It won't be hard to change if we pick wrong in any case.

iterators.tex
CaseyCarter: > Also, `__view_predicate` is a strange name. Why not `__is_view`?

Mostly to dissociate from the old trait and the connotations of naming something `is_foo`. `<type_traits>` has taught us that `is_foo` is a useful concept-membership trait; using a non-`is` name avoids the inevitable questions about turning it into a useful user-exposed trait. That said, I of course have no strong feelings about the naming of something that is `\expos`. Shall I change it?

ericniebler:
Merging this, but we need to revisit after the resolution of #71.

PR #77, opened 2015-08-02
Editorial fixes to DistanceType and iterator_distance_t

CaseyCarter:
Correct DistanceType to DifferenceType
Correct iterator_distance_t to DifferenceType


ericniebler:
Thanks

Issue #78, opened 2015-08-02
reverse_iterator


CaseyCarter:
The binary operator declarations have a lot of redundant constraints that could be simplified:
```
EqualityComparable{I1, I2}
bool operator==(const reverse_iterator<I1>& x,
                const reverse_iterator<I2>& y);

EqualityComparable{I1, I2}
bool operator!=(const reverse_iterator<I1>& x,
                const reverse_iterator<I2>& y);

TotallyOrdered{I1, I2}
bool operator<(const reverse_iterator<I1>& x,
               const reverse_iterator<I2>& y);

TotallyOrdered{I1, I2}
bool operator>(const reverse_iterator<I1>& x,
               const reverse_iterator<I2>& y);

TotallyOrdered{I1, I2}
bool operator<=(const reverse_iterator<I1>& x,
                const reverse_iterator<I2>& y);

TotallyOrdered{I1, I2}
bool operator>=(const reverse_iterator<I1>& x,
                const reverse_iterator<I2>& y);

SizedIteratorRange{I2, I1}
DifferenceType<I2> operator-(const reverse_iterator<I1>& x,
                             const reverse_iterator<I2>& y);

template <class I>
reverse_iterator<I> operator+(DifferenceType<I> n,
                              const reverse_iterator<I>& x);
```
since `reverse_iterator<T>` cannot be so much as referred to unless `T` satisfies the associated constraints of `reverse_iterator`, i.e., is a `BidirectionalIterator`.

* [reverse.iter.cons]/1:

  > *Effects:* Value initializes `current`. Iterator operations applied to the resulting iterator have defined behavior if and only if the corresponding operations are defined on a value-initialized iterator of type `I`. If `I` is a literal type, then this constructor shall be a trivial constructor.

  Since `I` is the type of `current`, it is not possible to satisfy both "value initializes `current`" and "If `I` is literal this constructor is trivial."

  I'm a huge fan of default initialization since I think it makes bugs more apparent than value-initialization, so I would favor declaring this constructor as explicitly defaulted. That said, I know most people don't agree with me so we should probably strike the sentence " If `I` is a literal type, then this constructor shall be a trivial constructor."

* [reverse.iter.op.star]:

  > *Effects:*
  ```
I tmp = current;
return *--tmp;
```

  could be simplified to "Effects: `return *prev(current);`

* [reverse.iter.opref]:

  > Returns: `current`.

  Should be "Returns: `prev(current)`"

ericniebler:
[...]

PR #79, opened 2015-08-02
reverse_iterator problems; Fixes ericniebler/stl2#78

CaseyCarter:
Default constructor cannot be trivial if it value-initializes `current`.
Correct off-by-one error in `operator->`.


CaseyCarter:
Same issue for `{back_,front_,""}insert_iterator`: a constructor cannot both value-initialize a subobject *and* be trivial.

Issue #71, opened 2015-07-29
Should Range<mutable_only_range>() be true or false?


ericniebler:
There's currently a problem in the Range and View concepts for a type like this:

```c++
struct mutable_only_range {
  int * begin();
  int * end();
};
```

`Range<mutable_only_range>()` is `false` because we need `begin(declval<mutable_only_range>())` to compile. Since there is no `begin` that takes rvalues, this selects the `begin(const T&)` overload, which tries to call `.begin()` on a const-qualified `mutable_only_range` object. Naturally, that fails.

Since `mutable_only_range` is not a `Range`, naturally it can't be a `View`, and it really seems like this type should satisfy `View`.

We could insist that `Range<mutable_only_range>()` is false but that `Range<mutable_only_range&>()` is true. That runs into trouble in `View` because reference types are not `Semiregular`. So both `View<mutable_only_range>()` and `View<mutable_only_range &>()` will be false. That obviously wrong.

The open questions are:

1. Is my intuition that this type is both a `Range` and a `View` correct?
2. If so, how do we change things to make that so? The options I see are:
  - Add an overload of `begin` that accepts rvalues.
  - Change the definition of `IteratorType` to be like the one in range-v3: if `T` has no top-level reference (lvalue or rvalue), a `&` is added. In this way, `View<mutable_only_range>()` is true but both `View<mutable_only_range&>()` and `View<mutable_only_range&&>()` are false.

(Obvious, the same is true for `end` and `SentinelType`.)

I like the range-v3 solution since it matches my expectation, so I'm going with that one for now.

ericniebler:
And yes, I'm aware that this is hackish and unprincipled and certain to surprise some people. We should seriously consider the other option of adding a `begin` that accepts rvalues. Arguments for and against? @CaseyCarter ?

CaseyCarter:
Yes, this is a bit of a mess. I'd prefer perfect-forwarding `begin` and `end` over having `IteratorType<T>` not be the same as `decltype(begin(declval<T>()))` - I think that way lies madness. My gut says that defining
```
template <class R>
concept bool View() {
  return Range<R>() && Semiregular<uncvref_t<R>>() && is_view<uncvref_t<R>>::value;
}
```
will correct the View problem, but I need to give it more thought.

ericniebler:
OK, please have a look at f25dc7a and tell me what you think.

ericniebler:
[...]

PR #81, opened 2015-08-04
Wording for issue ericniebler/stl2#71

CaseyCarter:
* Range, SizedRange, IteratorType, and SentinelType ignore ref-qualifiers
* Apply constraints for begin and end in the definitions of IteratorType and SentinelType.


iterators.tex
```diff
@@ -6327,19 +6327,26 @@
...
+  @\newtxt{\}}@
+using IteratorType = decltype(begin(declval<T@\newtxt{\&}@>()));
 
 template <class T>
-using SentinelType = decltype(end(declval<T>()));
+  @\newtxt{requires requires (T\& t) \{}@
+    @\newtxt{typename IteratorType<T>;}@
+    @\newtxt{\{ end(t) \} -> Sentinel<IteratorType<T>{}>; // not required to be equality preserving}@
+  @\newtxt{\}}@
+using SentinelType = decltype(end(declval<T@\newtxt{\&}@>()));
```
ericniebler: These definitions have the curious property that I can call `begin` on an object that doesn't have an `end`, but I can't call `end` on an object that doesn't have a `begin`. What's the rationale?

iterators.tex
```diff
@@ -6327,19 +6327,26 @@
...
-  @\newtxt{return}@ requires(T@\newtxt{\&\&}@ t) {
+  @\newtxt{return}@ requires@\oldtxt{(T t)}@ {
     typename IteratorType<T>;
     typename SentinelType<T>;
-    { begin(t) } -> @\newtxt{Same<}@IteratorType<T>@\newtxt{>}@; @\newtxt{// not \newtxt{required to be} equality preserving}@
-    { end(t) } -> @\newtxt{Same<}@SentinelType<T>@\newtxt{>}@; @\newtxt{// not \newtxt{required to be} equality preserving}@
-    requires Sentinel<SentinelType<T>, IteratorType<T>>@\newtxt{()}@;
+    @\oldtxt{\{ begin(t) \} -> IteratorType<T>;}@
+    @\oldtxt{\{ end(t) \} -> SentinelType<T>;}@
+    @\oldtxt{requires Sentinel<SentinelType<T>, IteratorType<T>{}>;}@
```
ericniebler: With the above definitions (which I'm iffy on), this could simply be "`return requires { typename SentinelType<I>; };`", right?

iterators.tex
```diff
@@ -6349,8 +6356,8 @@
 \oldtxt{\tcode{begin} and \tcode{end} are required to be amortized constant time.}
 
 \pnum
-\newtxt{Given an object \tcode{t} of type \tcode{T}, \oldtxt{the type \tcode{T} models}
-\tcode{Range\newtxt{<T>()}} \newtxt{is satisfied} if and only if}
+\newtxt{Given a value \tcode{t} of type \tcode{T}, \tcode{Range<T>()} is satisfied
+if and only if}
```
ericniebler: What's the rationale for changing `object` to `value`? Can you have a "value" of a reference type? I'm not even sure what that means.

ericniebler: Also, you have mismanaged the markup here. The change from the old "models" text to the new "satisfies" text has not been fully reviewed.

iterators.tex
CaseyCarter: The motivation is to minimally ensure that `IteratorType<T>` is always an `Iterator`, and `SentinelType<T>` is always a `Sentinel`. We could theoretically require both `begin` and `end` for both associated types, or even require `Sentinel<end(t), begin(t)>` for both, I went with minimal requirements for no particular reason.

iterators.tex
```diff
@@ -6391,14 +6398,14 @@
...
-      @\newtxt{\{ size(t) \} -> ConvertibleTo<DifferenceType<IteratorType<T\&>{}>{}>;}@
+      @\newtxt{\{ size(t) \} -> ConvertibleTo<DifferenceType<IteratorType<T>{}>{}>;}@
     @\newtxt{\};}@
 @\newtxt{\}}@
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
-\newtxt{Given an object \tcode{t} of type \tcode{T}, \tcode{SizedRange<T>()} is satisfied if and only if}
+\newtxt{Given a value \tcode{t} of type \tcode{T}, \tcode{SizedRange<T>()} is satisfied if and only if}
```
ericniebler: Ditto here about object -> value

iterators.tex
CaseyCarter: Yes - in fact it's implemented that way in cmcstl2. I included both in `Range` to provide an immediate visual link between the three things in the text. It's simply for readability.

iterators.tex
CaseyCarter: > What's the rationale for changing object to value? Can you have a "value" of a reference type?

You certainly can't have an object of a reference type. You're right, "value" isn't much better.

> Also, you have mismanaged the markup here.

I removed the nested markup: the whole sentence is `\newtxt`, I didn't think it necessary to have nested tags. The entire sentence is new since N4382, and to my knowledge hasn't received any review.

CaseyCarter: > "value" isn't much better.

How about "given an rvalue or lvalue `t` of type `T`, types `T`, `T&` and `T&&` model `Range` if and only if"?

CaseyCarter: ...but then we need to force `t` to an lvalue in the prose requirements...

CaseyCarter: "Given an lvalue `t` of type `T`, `Range<T>()`, `Range<T&>()`, and `Range<T&&>()` are satisfied if and only if"

CaseyCarter: That doesn't fix the `SizedRange` requirement: "Either `SizedRange<T>()` is satisfied, `T` does not meet...". I'll keep thinking.

CaseyCarter: How about ""Given an lvalue `t` of type `remove_reference_t<T>`, `Range<T>()` is satisfied if and only if"?

iterators.tex
ericniebler: It's not something I feel strongly about. The asymmetry is a little weird, that's all.

iterators.tex
ericniebler: OK

iterators.tex
ericniebler: That works, I think.

Also, I would strike the bullet about `SizedRange`. It doesn't seem appropriate to me. Are you trying to disallow `Ranges`s for which `size(r)` compiles but does the wrong thing? I think it's fine if such a type satisfies `Range` but not `SizedRange`.

CaseyCarter: > I think it's fine if such a type satisfies `Range` but not `SizedRange`.

We had a discussion about a thousand years ago in which you said that you want a template that is declared to require a `Range` to be able to assume that the actual parameter is a `SizedRange` if it meets the syntax. That's why I originally added this requirement that forbids passing a non-`SizedRange` that looks like a `SizedRange` as a `Range`.

If you've decided that you instead want to require anything that cares about `SizedRange`-ness to explicitly declare an overload/specialization for `SizedRange` - as you did with `distance` - then we can get rid of this bullet.

CaseyCarter: Why do we care that overload resolution selects "a unary non-member function"?

ericniebler: I honestly don't know. I stole the wording from the specification of `swap`.

\e

CaseyCarter:
There's a freshly rebased squashed commit with the "Satisfies `SizedRange` or..." requirement removed.

CaseyCarter:
Ok, note changed.

ericniebler:
Great, many thanks.

Issue #76, opened 2015-08-02
distance(Range)


CaseyCarter:
The proposal does not define `distance` for `Range`s, but only for `Iterator`/`Sentinel` pairs. Obviously for a `Range` `r` it is the case that `distance(r) == distance(begin(r), end(r))` so this isn't a huge problem.  However, for a model of `SizedRange` whose iterator and sentinel do not model `SizedIteratorRange`, `distance(begin(r), end(r))` has complexity `O(N)` whereas `distance(r)` is computable in `O(1)` with `size(r)`.

Is it worth the trouble of specifying `distance(Range)` for (a) typing convenience, (b) consistency with other `Iterator` + `Sentinel` algorithms that have a `Range` overload, and (c) better complexity for the above corner case? Off the top of my head, `container | view::move` for some container types models `SizedRange` but does not have random access iterators - I can't think of any other instance of this corner case.

ericniebler:
> Is it worth the trouble of specifying `distance(Range)` [...] ?

Yes.

`view::counted` is another example of a range that is sized but might not have random-access iterators. Also, a pair of `move_iterator`s (if we succeed in demoting it to single-pass).

PR #83, opened 2015-08-04
Wording for distance(Range). Fixes issue ericniebler/stl2#76.

CaseyCarter:
Wording itself is simple enough, organization is awkward. I inserted the para immediately after the wording for `distance(Iterator, Sentinel)` where I think it belongs. It's peculiar, however, that it appears before the range concepts are defined. In a perfect world, I would relocate the range concepts section before the `<iterator>` synoposis to correct the issue.


ericniebler:
Funny, I had just fixed this locally. I opted for a separate "Range primitives" section, right after "Range access".

PR #93, opened 2015-08-19
"undefined" -> "unspecified"

akrzemi1:
You do not want undefined behavior when the implementation decides not to check the additional constraints.

### DIFF
--- a/concepts.tex
+++ b/concepts.tex
@@ -138,7 +138,7 @@ struct T {
 \end{codeblock}
 
 Since \tcode{T} fails to satisfy the implicit requirements of \tcode{C}, \tcode{C<T>()} is not
-satisfied in principle. Whether it is satisfied in practice is undefined.
+satisfied in principle. Whether it is satisfied in practice is unspecified.
 \exitexample
 }
 


CaseyCarter:
This "whether it is satisfied in practice" wording is inconsistent with the current usage of "concept is satisfied" to mean that all requirements, both syntactic and pure semantic / prose, are met. It needs to be changed to something like:
> Since `T` fails to meet the implicit requirements of `C`, `C<T>()` is not satisfied. It is unspecified whether or not an implementation diagnoses a program that requires `C<T>()` as ill-formed.

And yes, we most certainly *do* want undefined behavior when someone instantiates a library template with parameters that meet the syntactically verified requirements but not the requirements that are not or cannot be verified. Given the definition of `C` generic code that requires `C<T>()` is free to assume that e.g. `(const T&)a == (const T&)b` is equivalent to `a == b` and act on that assumption. The compiler is free to optimize:
```
C{T}
void f(const T& a, const T& b) {
  if (a == b) {
    if ((T&)a == (T&)b) {
      // launch the missiles
    }
  }
}
```
into:
```
C{T}
void f(const T& a, const T& b) {
  if (a == b) {
    // launch the missiles
  }
}
```

PR #94, opened 2015-08-19
Removed casts to void

akrzemi1:
The casts are needed in Ranges v3, because the constrains are separated with commas. In C++ concepts, the constraints are separated by semicolons, so the cast to void adds nothing.
Also, the cast to void does not indicate that "we ignore the return value". It says "the result of swap should be castable to void", which is of no value.

### DIFF
--- a/concepts.tex
+++ b/concepts.tex
@@ -326,7 +326,7 @@ Then \oldtxt{types \tcode{T} and \tcode{U} model} \tcode{Assignable\newtxt{<T, U
 template <class T>
 concept bool Swappable() {
   return requires(T t, T u) {
-    @\newtxt{(void)}@swap(std::forward<T>(t), std::forward<T>(u));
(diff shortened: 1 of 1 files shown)


ericniebler:
Casey, the void casts greatly confused everybody on the telecon. We need to be more explicit about this notation.

CaseyCarter:
> Casey, the void casts greatly confused everybody on the telecon. We need to be more explicit about this notation.

I'll word a note. It being unclear to Andrzej was enough justification for me, LWG wondering why it's there is definite proof.

akrzemi1:
If I understood your intention correctly, it looks like rather than the casts to void, the document could specify a relaxation in [concepts.lib.general.equality], saying something like:
"In the cases where requirements in concepts do not constrain the return value of expressions, for the purpose of determining if these expressions are equality preserving or stable the expression's result is not considered its output."

Issue #95, opened 2015-08-19
Cast to void in the Swappable requirements is too subtle


CaseyCarter:
Needs an explanatory note to avoid confusion as in #94.

PR #98, opened 2015-08-24
Clarify the intent of the casts to void in Swappable's expressions.

CaseyCarter:
Fixes ericniebler/stl2#95. @akrzemi1, is the note text sufficient?


ericniebler:
Can't merge. Can you fix the conflicts?

ericniebler:
Although the void casts only show up in `Swappable` at the moment, it seems likely that they will appear elsewhere in the future. Perhaps the comment should be moved to the library introduction. We need more discussion there in general, particularly about what happens when instantiating a template with types that don't satisfy the templates constraints.

CaseyCarter:
> Although the `void` casts only show up in `Swappable` at the moment, it seems likely that they will appear elsewhere in the future.

My position is that the note is sufficient for now. If and when `void` casts spread we can go with [akrzemi1's suggestion](https://github.com/ericniebler/stl2/pull/94#issuecomment-132632942):

> the document could specify a relaxation in [concepts.lib.general.equality], saying something like:
"In the cases where requirements in concepts do not constrain the return value of expressions, for the purpose of determining if these expressions are equality preserving or stable the expression's result is not considered its output."

I don't want to introduce yet another implicit notational convention if we can a-`void` doing so: "implicitly equality preserving" and "implicit expression variants" is quite a bit for a reader to keep in mind already.

ericniebler:
> if we can a-`void` doing so

LOL. OK, it's fine, thanks.

PR #99, opened 2015-08-24
Correct "satisfied in principle" wording per discussion in …

CaseyCarter:
… ericniebler/stl2#93.


ericniebler:
Guessing the merge conflicts have to do with the pdf. I'm beginning to think that the pdf does not belong in the git repo.

concepts.tex
```diff
@@ -127,19 +127,23 @@
...
-Since \tcode{T} fails to satisfy the implicit requirements of \tcode{C}, \tcode{C<T>()} is not
-satisfied in principle. Whether it is satisfied in practice is undefined.
-\exitexample
+\oldtxt{Since }\tcode{T} fails to \oldtxt{satisfy}\newtxt{meet} the implicit
+requirements of \tcode{C}, so \tcode{C<T>()} is not satisfied\oldtxt{ in principle}.
+\oldtxt{Whether it is satisfied in practice is undefined.} \newtxt{Since
+implementations are not required to validate the syntax of implicit requirements, it
+is unspecified whether or not an implementation diagnoses a program which requires
+\tcode{C<T>()} as ill-formed.}
+ \exitexample
```
ericniebler: > "... a program which requires `C<T>()` as ill-formed"

That's a strange and awkward construction. "... a program that requires that `C<T>()` to be ill-formed"? And what does it mean that a concept check is ill-formed? That it might cause a hard error at compile time? I don't think that's what we want to say.

CaseyCarter: Is this clearer?

> Since implementations are not required to validate the syntax of implicit requirements, it is unspecified whether or not an implementation diagnoses as ill-formed a program which requires `C<T>()`.

ericniebler: Much.

PR #112, opened 2015-09-12
Specify constraints for tagged types

### LINKED ISSUE #92: tagged_{pair,tuple}
1) Should [tagged.pairs] become [tagged.pair] for symmetry with [tagged.tuple]?  (If so, then [tagged.pairs.creation] ought become [tagged.pair.creation]).

2) Should there be a synopsis subheading at the start of [tagged.pairs]?  I note that [tagged.tuple] has a synopsis subheading, but it is unnumbered and unlabeled. I recommend parallel treatment.

3) Notation such as `tag::min(int)` and `Tag1(V1)` is used several times in these sections, but I am puzzled by this syntax.  For example, section [alg.tagspec] specifies `tag::min` to be a struct with implementation-defined members, yet the parenthesis syntax suggests that it is instead some sort of callable that takes a type argument.  Can the intent of the parenthesized type be clarified, and/or alternate syntax provided?  Or what am I missing?

CaseyCarter:
Fixes #92.


ericniebler:
Could you update the pdf too?

CaseyCarter:
Done.

ericniebler:
Thanks

PR #114, opened 2015-09-15
Cleanup references

CaseyCarter:
* Add proper bib for N1550.
* Remove bad rows from the "Ranges library summary" table in [ranges.general]/2, add row for [iterator.range]
* Change tag of the "Ranges TS library headers" table to `tab:intro.headers` instead of duplicating `tab:iterators.relations`


ericniebler:
Thanks. FYI, I'm on vacation with my family this week so I'll be somewhat less responsive that normal.

PR #115, opened 2015-09-16
Specify OutputRange

CaseyCarter:
Since `fill` and `generate` require `OutputRange`, we should probably specify what it is ;)


iterators.tex
```diff
@@ -6669,6 +6669,22 @@
...
+
+\pnum
+The \tcode{OutputRange} concept describes requirements of an \tcode{Range} type
+for which \tcode{begin} returns a type that satisfies
+\tcode{OutputIterator}~(\ref{output.iterators}).
+
+\begin{codeblock}
+template <class T>
+concept bool OutputRange() {
+  return Range<T>() && OutputIterator<IteratorType<T>>();
```
ericniebler: `OutputIterator` is a two-argument concept, is it not?

CaseyCarter: Yes, of course - as should be obvious to anyone who is paying attention. Fixed and rebased onto master.

CaseyCarter: ...except you changed master again. re-rebased.

PR #125, opened 2015-09-24
Minor editorial fixes in algorithm declarations

CaseyCarter:
I don't think the spelling fix is deserving of change tags.


ericniebler:
Thanks

PR #143, opened 2015-11-05
Specify N4381-style customization point objects for {c,r}begin, {c,r}…

CaseyCarter:
…end, size, empty, {c}data, and swap.

I still need to:

* update references to the customization points in the rest of the TS to make sure they agree with the specified semantics (especially the discussion of `Swappable`) and probably qualify them a la `ranges::begin` to make it clear that unqualified name lookup isn't happening.

* Fix declarations of functions with the same name as a customization point, e.g. `template <....> void swap(tagged<...>&, tagged<...>&)`.


concepts.tex
```diff
@@ -404,17 +409,17 @@
...
+                                                        // for rvalues and lvalues
 }
 
 // Requires: lvalues of \tcode{T} shall be swappable.
 template <class T>
 void lv_swap(T& t1, T& t2) {
-  using std::swap;
-  swap(t1, t2);                                 // OK: uses swappable conditions for
-}                                               // lvalues of type \tcode{T}
+  @\oldtxt{using std::swap;}@
```
ericniebler: I use `oldoldtxt` when changing official working draft wording so I know what parts should end up in red strikethrough. It makes my job as editor much easier.

iterators.tex
```diff
@@ -7346,23 +7352,287 @@
...
 \end{itemdescr}
+}
+
+{\color{newclr}
+\rSec2[iterator.range.begin]{\tcode{begin}}
+\pnum
+The name \tcode{begin} denotes a customization point object
+(\ref{customization.point.object}). The expression
+\tcode{ranges::begin(E)} for a subexpression \tcode{E} has the same
+effects as:
```
ericniebler: "The effect of the expression `ranges::begin(E)` for some expression `E` is equivalent to:"

iterators.tex
```diff
@@ -7346,23 +7352,287 @@
...
+
+\begin{itemize}
+\item
+  \tcode{(E).begin()} if \tcode{E} has type
+  \tcode{std::initializer_list<T>} for some type \tcode{T}.
```
ericniebler: What if it has type `std::initializer_list<T> &`?

iterators.tex
```diff
@@ -7346,23 +7352,287 @@
...
+
+\item
+  Otherwise, \tcode{ranges::begin((const T\&)(E))} if \tcode{E} is an
+  rvalue of some type \tcode{T}. This usage is deprecated.
```
ericniebler: ~~some~~

iterators.tex
```diff
@@ -7346,23 +7352,287 @@
...
+  Otherwise, \tcode{(E) + 0} if \tcode{E} has array type.
+
+\item
+  Otherwise, \tcode{(E).begin()} if it is a valid expression whose
+  type \tcode{I} meets the syntactic requirements of \tcode{Iterator<I>}.
+  If \tcode{Iterator} is not satisfied, the program is ill-formed
+  with no diagnostic required.
+
+\item
+  Otherwise, \tcode{begin(E)} if it is a valid expression whose type
```
ericniebler: "if it is a valid expression" -> would probably be covered by the blanket effects-equivalent-to above.

iterators.tex
```diff
@@ -7346,23 +7352,287 @@
...
+  \tcode{I} meets the syntactic requirements of \tcode{Iterator<I>} when
+  overload resolution is performed in a context that includes the
+  declaration \tcode{void begin(auto\&) = delete;} and does not include
+  a declaration of \tcode{ranges::begin}. If \tcode{Iterator}
+  is not satisfied, the program is ill-formed with no diagnostic
+  required.
+
+\item
+  Otherwise, \tcode{ranges::begin(E)} is not a valid expression and
+  shall result in a substition failure.
```
ericniebler: "shall result in substitution failure" -> I find no mentions of substitution failure in the standard. I think it's enough to simply say that `begin(E)` is ill-formed.

concepts.tex
CaseyCarter: I've already changed these to `using std::@\newtxt{experimental::ranges::}@swap;` since `ranges::swap` is not the full name of `std::experimental::ranges::swap`, i.e., I broke the examples.

iterators.tex
```diff
@@ -7346,23 +7352,287 @@
...
+
+\item
+  Otherwise, \tcode{ranges::begin((const T\&)(E))} if \tcode{E} is an
+  rvalue of some type \tcode{T}. This usage is deprecated.
+  \enternote This deprecated usage exists so that
+  \tcode{ranges::begin(E)} behaves similarly to \tcode{std::begin(E)}
+  as defined in C++14 when \tcode{E} is an rvalue. \exitnote
+
+\item
+  Otherwise, \tcode{(E) + 0} if \tcode{E} has array type.
```
ericniebler: "array type~(\ref{basic.compound})". Also, what about references to arrays?

iterators.tex
CaseyCarter: Expressions never have reference type; that would be "is an lvalue of type `std::initializer_list<T>`." This case doesn't care about value category.

iterators.tex
ericniebler: Actually, I think once you strike "if it is a valid expression" for the bullet above, this bullet becomes unnecessary.

iterators.tex
```diff
@@ -7346,23 +7352,287 @@
...
+  as defined in C++14 when \tcode{E} is an rvalue. \exitnote
+
+\item
+  Otherwise, \tcode{make_reverse_iterator((E) + 0)} if \tcode{E} has
+  array type.
+
+\item
+  Otherwise, \tcode{(E).rend()} if it is a valid expression whose
+  type \tcode{S} meets the syntactic requirements of
+  \tcode{Sentinel<S, decltype(ranges::rbegin(E))>}. If
```
ericniebler: you need empty `()` after `Sentinel<S,I>` consistently throughout this patch.

iterators.tex
```diff
@@ -7346,23 +7352,287 @@
...
+  \tcode{Sentinel} is not satisfied, the program is ill-formed with
+  no diagnostic required.
+
+\item
+  Otherwise, \tcode{make_reverse_iterator(ranges::begin(E))} if both
+  \tcode{ranges::begin(E)} and \tcode{ranges::end(E)} are valid
+  expressions of the same type \tcode{I} which meets the syntactic
+  requirements of \tcode{BidirectionalIterator}. If
+  \tcode{BidirectionalIterator<I>()} is not satisfied, the program is
```
ericniebler: \ref{iterators.bidirectional}

iterators.tex
CaseyCarter: Done.

iterators.tex
CaseyCarter: Done.

iterators.tex
```diff
@@ -7402,6 +7674,130 @@
...
+  resolution is performed in a context that includes the declaration
+  \tcode{void size(const auto\&) = delete;} and does not include
+  a declaration of \tcode{ranges::size}.
+
+\item
+  Otherwise,
+  \tcode{ranges::cend(E) - ranges::cbegin(E)}, except that \tcode{E}
+  is only evaluated once, if it is a valid expression and the types
+  of \tcode{ranges::cbegin(E)} and \tcode{ranges::cend(E)} meet the
+  syntactic requirements of \tcode{SizedIteratorRange}. If
```
ericniebler: \ref{iteratorranges.sizediteratorrange}. Need to say something here about the signed-ness of the returned integral type, though. I think in the current draft, this needs to be coerced to an unsigned type.

utilities.tex
```diff
@@ -5,36 +5,27 @@
...
 Change the \tcode{<utility>} synopsis as follows:
 
 \begin{codeblock}
-@\added{namespace std \{}@
   // ... as before ..
 
   // \ref{utility.swap}, swap:
   template<class T>
-    @\added{requires experimental::ranges\oldtxt{_v1}::Movable<T>()}@
+    @\oldtxt{requires experimental::ranges\oldtxt{_v1}::Movable<T>()}@
```
ericniebler: If you're striking this whole section, then just delete these lines instead of marking them as deleted.

utilities.tex
```diff
@@ -5,36 +5,27 @@
...
-\ednote{``\tcode{swap}'' is a customization point found by Argument-Dependent Lookup. Due to the nature of
-ADL, the \tcode{swap} in namespace \tcode{std} would be considered if either argument has namespace \tcode{std}
-as an associated type. If \tcode{std::swap} remains unconstrained, then the \tcode{Swappable} concept will
-erroneously return true for some types that are not in fact swappable. Below, we suggest adding constraints
-directly to the overloads of \tcode{swap} in namespace \tcode{std} rather than defining new constrained
-overloads of \tcode{swap} in namespace \tcode{std::experimental::ranges\oldtxt{_v1}}. The precise mechanism by which
-\tcode{swap} is constrained such that it is usable in build environments lacking the concepts language feature
-is unspecified.}
-
+{\color{oldclr}
```
ericniebler: oldoldclr

utilities.tex
```diff
@@ -124,13 +118,69 @@
...
+\begin{itemize}
+\item
+  \tcode{(void)swap(E1, E2)} if that is a valid expression when
+  overload resolution is performed in a context that includes
+  the declarations
+\begin{codeblock}
+  template <class T>
+  void swap(T&, T&) = delete;
+  template <class T, size_t N>
+  void swap(T(&)[N], T(&)[N]) = delete;
```
ericniebler: I think the purpose of these `= delete` poison pill declarations is obscure enough to get a Note.

utilities.tex
```diff
@@ -124,13 +118,69 @@
...
+\pnum
+\remarks Whenever \tcode{ranges::swap(E1, E2} is a valid
+expression, it exchanges the values denoted by \tcode{E1}
+and \tcode{E2} and has type \tcode{void}.
+}
+
+\ednote{\oldtxt{The following edits to the two \tcode{swap} overloads are to be applied to the versions in
+namespace \tcode{std}.}}
+
+{\color{oldclr}
```
ericniebler: oldoldclr, right?

utilities.tex
```diff
@@ -124,13 +118,69 @@
...
 \begin{itemdecl}
 template<class T>
-  @\added{requires experimental::ranges\oldtxt{_v1}::Movable<T>()}@
+  @\oldtxt{requires experimental::ranges\oldtxt{_v1}::Movable<T>()}@
```
ericniebler: delete

iterators.tex
ericniebler: OK, but I still think this bullet can be safely dropped, no?

ericniebler:
FYI, I'm liking this direction.

iterators.tex
CaseyCarter: Done.

iterators.tex
CaseyCarter: and done.

iterators.tex
CaseyCarter: I specifically want to exclude initializer_list rvalues from the general deprecation of rvalues in the next bullet.

ericniebler: Why? It's equally unsafe, is it not? Backing storage for an `initializer_list` is on the stack IIUC.

iterators.tex
CaseyCarter: Done.

iterators.tex
CaseyCarter: Do you have a problem with me abbreviating to simply the concept name in the second occurrence below?

iterators.tex
CaseyCarter: Done.

iterators.tex
CaseyCarter: `SizedRange` only requires that `size()` returns a model of `Integral` that is implicitly convertible to `difference_type_t<iterator_t<Range>>`, i.e., it does not require `UnsignedIntegral`.

utilities.tex
CaseyCarter: Woops - think I need to fix this in `Swappable` as well.

utilities.tex
CaseyCarter: Done.

utilities.tex
CaseyCarter: Yes, thanks - I'd thought about this but forgotten to write one.

utilities.tex
CaseyCarter: Right, there are several in this section.

utilities.tex
CaseyCarter: Done.

iterators.tex
CaseyCarter: > Why?

Compatibility with the WP, I thought, but it doesn't have `initializer_list` overloads for `begin` and `end`. In any case it's a horrible idea for exactly the reasons you state, so let's deprecate it.

CaseyCarter: > it doesn't have `initializer_list` overloads for `begin` and `end`

I'm not completely insane - there are overloads of `rbegin` and `rend` that take `initializer_list` by value. They can go as well.

iterators.tex
ericniebler: No.

iterators.tex
ericniebler: OK. IMO, we should eventually banish unsigned integer types from STL2, but now is not the time.

utilities.tex
CaseyCarter: There's a fresh push with all the discussed changes and a perhaps overly-elaborate discussion/example of the poison pill technique in lib-intro. I still have a few more things to clean up on my end, so don't merge this yet. Feel free to point out any more stupid things I've done ;)

CaseyCarter:
I now have everything in here that I want, so we can merge as soon as you're happy with it.

> FYI, I'm liking this direction.

Yeah, kudos for getting me to raise the level of abstraction in the specification of the customization points - they're much cleaner.

iterators.tex
```diff
@@ -1702,32 +1702,48 @@
...
 \begin{addedblock}
 \begin{codeblock}
-  template <@\oldtxt{class}\newtxt{_Auto}@ C> auto size(const C& c) -> decltype(c.size());
-  template <@\oldtxt{class}\newtxt{_Auto}@ T, size_t N> constexpr size_t size(T (&array)[N]) noexcept;
-  template <@\oldtxt{class}\newtxt{_Auto}@ E> @\newtxt{constexpr}@ size_t size(initializer_list<E> il) noexcept;
+  @\oldtxt{template <class C> auto size(const C\& c) -> decltype(c.size());}@
+  @\oldtxt{template <class T, size_t N> constexpr size_t size(T (\&array)[N]) noexcept;}@
+  @\oldtxt{template <class E> size_t size(initializer_list<E> il) noexcept;}@
+  @\newtxt{namespace \{}@
+    @\newtxt{constexpr \impdef{} begin;}@
```
ericniebler: Can constexpr objects be fwd-declared like this? I don't think so. We may need an `= \impdef;` here. And do we really want "implementation defined" instead of unspecified?

CaseyCarter: No, they can't. `constexpr unspecified begin = unspecified;` is enough to guarantee constant initialization, albeit a bit unattractive.

lib-intro.tex
```diff
@@ -72,3 +72,124 @@
...
+\rSec4[customization.point.object]{Customization Point Objects}
+\pnum
+A \techterm{customization point object} is an object with a
+literal class type which interacts with user-defined types while
+enforcing semantic requirements on that interaction.
+
+\pnum
+If \tcode{T} is the type of a customization point object, \tcode{T}
+shall satisfy \tcode{Semiregular<T>()} and
+\tcode{Function<const T, Args..>()} when the types of \tcode{Args..}
```
ericniebler: ... and the type `result_of_t<const T&&(Args...)>`. Think of `begin` being constrained to only return `Iterator`s.

lib-intro.tex
```diff
@@ -72,3 +72,124 @@
...
+namespace \tcode{std}~(C++ \cxxref{utility.swap}):
+\begin{codeblock}
+template <class T>
+void swap(T&, T&) noexcept(is_nothrow_move_constructible<T>::value &&
+                           is_nothrow_move_assignable<T>::value);
+template <class T, size_t N>
+void swap(T(&a)[N], T(&b)[N]) noexcept(noexcept(swap(*a, *b)));
+\end{codeblock}
+\tcode{ranges::swap} specifies that the overload resolution
+it performs internally for \tcode{swap(E1, E2)} for some
```
ericniebler: s/it/is/

lib-intro.tex
```diff
@@ -72,3 +72,124 @@
...
+result in \#1 being called, since it is more specialized than the
+``poison pill'' overload \tcode{swap(T\&, T\&)}. A call with lvalues
+of type \tcode{B<int>} and \tcode{B<double>} will invoke \#2, since
+the poison pill overload is not viable. A call with two lvalues of
+type \tcode{B<int>}, however, will fail overload resolution since
+both \#2 and the poison pill overload are viable and neither is
+more specialized nor more constrained.
+
+A call with lvalues of type \tcode{D<int>} and \tcode{D<int>}
+results in a lookup set containing \#3, the poisin pill overload,
```
ericniebler: s/poisin/poison/

lib-intro.tex
```diff
@@ -72,3 +72,124 @@
...
+and \tcode{std::swap(T\&, T\&)} since \tcode{D} inherits from
+\tcode{std::string} making \tcode{std} an associated namespace.
+Overload resolution will unambiguously select \#3 since it
+is more constrained than the other two overloads.
+\exitexample
+}
```
ericniebler: We will certainly get dinged for this long discourse. The Standard Is Not A Tutorial&#8482;. It feels like this belongs in an appendix, and the Working Paper itself should have a \note that says something simple like, "The purpose of the `delete`ed overload(s) is to prevent the unconstrained overloads in namespace `std` from being selected by overload resolution."

utilities.tex
```diff
@@ -124,13 +116,67 @@
...
+  exchange the values denoted by \tcode{E1} and \tcode{E2},
+  the program is ill-formed with no diagnostic required.
+
+\item
+  Otherwise, \tcode{(void)swap_ranges(E1, E2)} if \tcode{E1} and
+  \tcode{E2} are lvalues of array types~(\cxxref{basic.compound})
+  of equal extent and \tcode{ranges::swap(*(E1), *(E2))}
+  is a valid expression, except that
+  \tcode{noexcept(ranges::swap(E1, E2))} is equal to
+  \tcode{noexcept(ranges::swap(*(E1), *(E2)))}.
```
ericniebler: You have generalized this to include swapping arrays of different type. I know that's your ultimate intent, but perhaps this deserves an \ednote stating that this generalization was intentional. It deserves review discussion.

lib-intro.tex
CaseyCarter: If I'm understanding, your suggestion here is that each customization point object type should specify its results to satisfy a specified concept - as all those currently specified do if we call `Same<void>` the result concept of `swap` - and that this should be a required feature of \techterm{customization point object}? (If I'm misunderstanding and that's *not* your suggestion, I may put it in anyway because it's a great idea.)

ericniebler: Yes, that's what i was saying. The customization point will probably constrain its return type in addition to its arguments, and that should be part of the object's contract.

CaseyCarter: "Each customization point object type constrains the type of its results to satisfy a particular concept."?

ericniebler: Some customization points might not constrain their return type, but others might.

lib-intro.tex
CaseyCarter: When I started thinking about introducing "poison pill" as a term-of-art, I noticed that this "example" is the majority of the section and I knew I had gone off the rails. I want to make two points here: (1) the deleted overloads prevent calling `std` functions, and (2) user overloads necessarily must be more specialized/more constrained than the deleted overloads to hook the customization point when the deleted overloads are viable. (2) has a subtle danger:

```c++
template <class T> struct A {};
template <class T, class U> void swap(A<T>&, A<U>&) {}
```
If I call `ranges::swap` with lvalues `A<int>` and `A<double>`, my swap overload is invoked. If I call with `A<int>` and `A<int>` it is not invoked, and I don't get a compile error - `Movable<A>()` is satisfied so the default swap implementation is used.

How about:
```tex
\enternote Many of the customization points objects in the library
evaluate function call expressions with an unqualified name which
results in a call to a user-defined function found by argument
dependent name lookup~(\cxxref{basic.lookup.argdep}). To preclude
such an expression resulting in a call to unconstrained functions
with the same name in namespace \tcode{std}, customization point
objects specify that lookup for these expressions is performed in
a context that includes deleted overloads matching the signatures
of overloads defined in namespace \tcode{std}. When the deleted
overloads are viable, user-defined overloads must be more
specialized~(\cxxref{temp.func.order}) or more
constrained~(N4377 [temp.constr.order])  to be used by a
customization point object. \exitnote
```

ericniebler: Good.

ericniebler: I wonder if there is a technical fix for this. `ranges::swap` could in theory test for the well-formed-ness of (unqualified) `swap(a,b)` when `a` and `b` have different types. The unconstrained overloads in `std` won't be considered because they require that `a` and `b` have the *same* type. `ranges::swap` could then prefer that to the general overload constrained with `Movable`. Thoughts?

CaseyCarter: *sigh* I'm rushing to get this finished, and not thinking things through. I wasted an hour trying to implement this before realizing that it won't help because the problem case is when they have the *same* type, e.g., `A<int>` and `A<int>`. If the types are different then the poison pill (template<class T> void swap(T&, T&) = delete` isn't viable anyway.

utilities.tex
CaseyCarter:
```tex
\ednote{This formulation intentionally allows swapping arrays
with identical extent and differing element types, but only
when swapping the element types is well-defined. Swapping
arrays of \tcode{int} and \tcode{double} continues to be
unsound, but \tcode{Swappable<T\&, U\&>()} implies
\tcode{Swappable<T(\&)[N], U(\&)[N]>()}.}
```

CaseyCarter:
Force-pushed a squashed-and-rebased commit.

ericniebler:
Merged. Woohoo!

CaseyCarter:
Goodbye `using std::swap; swap(...);`, we'll miss you.

PR #148, opened 2015-11-09
Add para [alg.general]/16 requiring implementations to resolve

### LINKED ISSUE #145: Iterator/Range Ambiguity
Any pair of overloads that differs in a single argument such that one overload requires `FooIterator` and the other `BarRange` needs disambiguation - arrays are both (undecayed) ranges and (decayed) iterators. Arrays make better ranges than iterators due to being statically sized, so we want the range overload to be preferred.

- [x] Determine the best way to specify the disambiguation, whether it be explicitly specifying additional requirements, or an implicit blanket requirement that any such occurrences of ambiguity must be resolved in favor of the range overload when applicable.
- [x] If the decision is to explicitly specify requirements, audit the algorithms for all such occurrences.
- [x] Write wording for the chosen disambiguation solution.

CaseyCarter:
… range/iterator ambiguity without specifying a particular mechanism. Fixes #145.


algorithms.tex
```diff
@@ -1412,6 +1412,15 @@
...
 iterator and sentinel arguments.
+
+\pnum
+\newtxt{Some algorithms declare both an overload that takes a \tcode{Range}
+and an \tcode{Iterator}, and an overload that takes two \tcode{Range}
+arguments. Since an array type~(\cxxref{basic.compound}) both satisifies
+\tcode{Range} and decays to a pointer~(\cxxref{conv.array}) which satisfies
+\tcode{Iterator}, such overloads are ambiguous when an array is passed as
+the second argument. Implementations shall provide a mechanism to resolve
+this ambiguity in favor of the overload that takes two ranges.}
```
ericniebler: I think we also need a paragraph saying what part of the algorithm signatures can be relied upon and which cannot. Users are already prohibited from taking the address of `std` functions. We also need to say that the order of template parameters is unspecified, for instance. Whether an iterator or range is passed by value or by forwarding reference is technically observable -- the iterator's copy constructor might have side-effects -- so we need to be clear that they can't rely on iterators being copied into algorithms just because the signature that appears in the Ranges TS happens to give the impression that iterators will be taken by value.

ericniebler: > We also need to say that the order of template parameters is unspecified, for instance.

This could make lots of code UB. `std::min<int>(x,y)` is commonplace, for instance. We might want blanket wording about the order of template parameters, and then make an exception for this case, but I'm not sure.

CaseyCarter: > Whether an iterator or range is passed by value or by forwarding reference is technically observable

I've been operating under the assumption that given unordered template parameters it's unobservable whether
function parameters are taken by value or forwarding reference. That's only true, however, if the initialization of copies from the arguments is sequenced before any invocation of user code. That is clearly a ridiculous standard to require of implementations; certainly mine doesn't meet it. Every one of my range algorithms, for example, calls `begin` and `end` on the range and perfect-forwards the other arguments into an iterator algorithm where they will eventually be copied/moved. I would rather have fewer total copies/moves than have the first set of copies/moves occur in the proper sequence.

I wonder if we'd be better served by raising the level of semantic abstraction for the algorithm specifications as we did for customization points. Instead of specifying the existence and behavior of two overloaded function templates named, e.g., `find_if`, we would specify the effects of the expression `ranges::find_if(i, s, t)` and `ranges::find_if(r, t)` when the types of named parameters satisfy the appropriate concepts. Template parameters and any ordering thereof never enter into the picture. That style specification would even allow - but not require - implementation of `find_if` as an object instead of a set of function template overloads.

CaseyCarter: > That style of specification would even allow - but not require - implementation of `find_if` as an object instead of a set of function template overloads.

I'm really stretching here. The absence of ADL alone would break an implementation using objects instead of function templates. Let's just ignore that sentence for now.

ericniebler: What you're describing here is a sweeping change to the specification of all the algorithms. However well intentioned that is, it's a larger change than I am comfortable with at this point. I prefer to keep things as they are, with some blanket wording about template parameters and function signatures.

CaseyCarter: Yes, I've been pacing about the house considering pros and cons and have come to the same conclusion. Six months ago this would have been a great idea, but it's too much of a design change this late in the process. The change to the customization points was absolutely necessary to give them well-defined semantics; that is not the case for the algorithms.

asutton: >
> Yes, I've been pacing about the house considering pros and cons and have come to the same conclusion. Six months ago this would have been a great idea, but it's too much of a design change this late in the process. The change to the customization points was absolutely necessary to give them well-defined semantics; that is not the case for the algorithms.
>
> FWIW, this is what I've thought is the right way to document generic
algorithms for a long time, but I didn't know if would scale to the
complexity of the standard. It would almost certainly be worth discussion
in LWG.

I wouldn't try to apply it to the TS in flight. It might be handled
editorially after Jacksonville -- assuming people agree with the editorial
change.

ericniebler: I tend to doubt a change of this magnitude could plausibly be considered editorial.

CaseyCarter:
New pull with wording for unspecified order and number of template parameters. I'm inclined not to special case `min`/`max`/`minmax` as discussed herein due to the recent discussion on the reflectors about fixing those algorithms in regards to rvalue arguments.

CaseyCarter:
> Whether an iterator or range is passed by value or by forwarding reference is technically observable...

I still need to fix this by making it unspecified if and when the actual argument expressions are copied.

CaseyCarter:
Ok, I believe I've addressed all the concerns discussed herein and this is ready to merge.

algorithms.tex
```diff
@@ -1412,6 +1412,36 @@
...
+
+\pnum
+\newtxt{The number and order of template parameters for algorithm declarations
+is unspecified, except where explicitly stated otherwise. The nominal parameters
+specified herein for each function template declaration are only significant in
+that they specify the association of template constraints with function
+parameters.}
```
ericniebler: I would strike the second sentence. The first one says it all.

CaseyCarter: Works for me.

algorithms.tex
```diff
@@ -1412,6 +1412,36 @@
...
+initialize the actual parameters except that any such initialization shall be
+sequenced before~(\cxxref{intro.execution}) the algorithm returns.
+\enternote The behavior of a program that modifies the values of the actual
+argument expressions is consequently undefined unless the algorithm return
+happens before~(\cxxref{intro.multithread}) any such modifications. \exitnote}
+
+\ednote{The intent is to allow algorithm implementations to accept parameters
+by forwarding reference and delay or omit initialization of the actual
+parameter, either as an optimization or to implement the array disambiguation
+requirement described above.}
```
ericniebler: I know this is what we discussed, but reading it gives me a "yuk" feeling. I hate introducing UB here. If we take the long-term view -- where the algorithms are specified purely by the semantics of expressions like `find_if(i, s, pred{, proj})` -- then how would we specify it? I guess it would be a lot like the above, wouldn't it? Blarg.

CaseyCarter: Every requirement in the specification implicitly introduces UB, at least this one spells it out clearly. The algorithm will probably initialize new objects from your expressions, but it may call back into your code before and/or after doing so, so don't do anything stupid. Otherwise something as simple as:
```c++
bool any_of(Iterator i, Sentinel<Iterator> s, XXX&& pred_, YYY&& proj_) {
  if (i != s) {
    auto pred = foo(pred_);
    auto proj = foo(proj_);
    // ...
  }
  return true;
}
```
isn't possible under as-if, since (a) the call of a possibly user-supplied `!=` happens before pred and proj are initialized, and (b) pred and proj may not be initialized *at all*, which is definitely observable.

ericniebler:
OK, thanks.

PR #149, opened 2015-11-13
Function becomes Callable:

CaseyCarter:
* Import `invoke` from the WP. I've numbered its section 20.9.3 as in the WP, which could cause confusion since 20.9.3 is [refwrap] in C++14.
* Despite that they are equivalent, replace occurrences of *`INVOKE`* with `invoke` because *`INVOKE`* is hard on the eyes.
* Rename all sections in concepts.lib.functions to concepts.lib.callables. Retitle the section "Callable concepts".
* Rename concept `Function` to `Callable`, change its required expression to `invoke(f, args...)`.
* Rename concept `RegularFunction` to `RegularCallable`.
* Remove section indirectcallables.as_function_t, specify the `IndirectCallableFoo<F, ...Is>` concepts to require `Foo<F, value_type_t<Is>...>`.
* Update `indirect_result_of` and `projected` appropriately.
* Require the type of a customization point object to satisfy `Callable` instead of `Function`. This is awkward in that it seems to imply that CP objects need to be passed to `invoke`, although the individual specifications make it clear that is not the case.
* Cleanup the specification of the `generate` algorithm family. (I had to touch it because of the `invoke` change, and thought it was overcomplicated. `fill` could use a similar treatment.)

Drive-by fixes:
* Fix occurrence of type/reference confusion in the specification of `indirect_result_of`. (We don't want to constrain e.g. `less<>&` to satisfy `IndirectCallable`, we want to constrain `less<>`.)
* Simplify concept `UniformRandomNumberGenerator`, reorder requirements for `shuffle` properly.


ericniebler:
Awesome. Fix those two minor issues and the rest looks good.

CaseyCarter:
Corrected review issues, and hopefully cleared up any customization point object confusion by explicitly specifying that a customization point object is a function object.

PR #150, opened 2015-11-14
Strong semantics for Assignable.

CaseyCarter:
* Remove the default for `Assignable`'s second parameter.
* Remove prose requirements from `Copyable` and `Movable`, they are now provided by `Assignable`.
* `exchange` and default-`swap` both now require `MoveConstructible` and `Assignable` instead of `Movable`.
* `Movable` can now explicitly require `Swappable`, and all is right with the world.
* Use `Assignable` instead of `is_assignable` in the specification of `advance(I, S)` to ensure the post-condition (`i == s`).


CaseyCarter:
There's a second commit that simplifies the specification of `advance(I, S)` which became too complicated post-Kona. My `Assignable` changes make the requirements worse, or at least no better, this change is crystal clear.

ericniebler:
I'm not sure how I feel about this. Think of output iterators and proxy references. I'm what sense is some value assignable to *i when i is an ostream iterator? I feel like Assignable isn't the right place to put these semantics.

CaseyCarter:
We don't use `Assignable` in `Writable` and `MoveWritable` for exactly that reason; the semantics are different.

CaseyCarter:
Integrating these semantics for `Assignable` consistently with iterators would be in scope for https://github.com/CaseyCarter/stl2/issues/17, and definitely something I'll be thinking about at length.

CaseyCarter:
There's a fresh pull with `Assignable` requiring `Common`. I've yanked the `advance` changes from the branch since they are a separate issue, and I'm working on a more comprehensive cleanup of `Sentinel` / `SizedIteratorRange` / `advance`.

ericniebler:
I'm not seeing the `Common` requirement of `Assignable`. Did you forget to add it?

CaseyCarter:
> I'm not seeing the Common requirement of Assignable. Did you forget to add it?

I reset the branch, and ran `git commit --amend` without adding the changes to the index. Yay me.

Issue #151, opened 2015-11-17
`std::less<T*>` is ill-formed.


tomaszkam:
According to the current version of the Range TS, the `std::less<T>` is requiring that `T` should should satisfy `StrictTotalOrder` or be same as `void`. Notice that both of this constrain are not satisfied if `T` is an pointer type, as `operator<` is required to represent only partial (points to same array) order on pointer  types, and this breaks semantics for `StrictTotalOrder` concept. According to 17.5.1.3 [structure.requirements] p9 this makes programs using `std::less` with pointer type are ill-formed.: 
> If the semantic requirements of a declaration are not satisfied at the point of use, the program is ill-formed, no diagnostic required.

This is esentailly contradiction to initial purpose of introduction of `std::less` wrappers that was designed to provide an total order for poitner types. The resolution may be to change the requirement for `std::less` (and others) to require that `T` is either `void`, pointer or models `StrictTotalOrder`,

asutton:
Maybe we should define the comparators as:

template<typename T>
struct less;

template<>
struct less<void> { ... };

template<typename T>
struct less<T*> { ... };

template<TotallyOrdered T>
struct less<T> { ... };

If a top-level constraint is needed:

template<typename T>
  requires is_void_c<T> || is_pointer_c<T> || TotallyOrdered<T>()
struct less;

Typical usage means you're going to instantiate the comparator immediately,
so you wouldn't get bad errors messages --- actually they might be better
(no specialization found, candidates are...)


Andrew Sutton

CaseyCarter:
There should definitely *not* be a top-level constraint on the template. There's no need to constrain which types may have an ordering defined for them, all we need do is constrain the specializations to apply only to types they are capable of ordering. The effect of the top-level constraint is to forbid me to specialize `less` for `Widget` unless `Widget` is `StrictTotallyOrdered`, which is ridiculous. If that were the case I wouldn't need to specialize `less` in the first place.

It would be convenient if user specializations of `less` were required to satisfy `StrictWeakOrder` - as the library-supplied implementations do - so that generic code that requires an ordering for `T` is guaranteed that `less<T>` provides one if it is defined.

ericniebler:
Agree with all of this. I'll make the change.

asutton:
>
> There should definitely *not* be a top-level constraint on the template.
> There's no need to constrain which types may have an ordering defined for
> them, all we need do is constrain the specializations to apply only to
> types they are capable of ordering.
>
Good point.


[...]

PR #152, opened 2015-11-18
respecify the comparison function objects

ericniebler:
Refs #151 . @CaseyCarter rip it apart.


utilities.tex
```diff
@@ -276,28 +276,45 @@
...
+  struct less_equal; @\newtxt{// not defined}@
+  @\newtxt{template<StrictTotallyOrdered T> struct less_equal<T>;}@
+  @\newtxt{template<class T> struct less_equal<T*>;}@
+
+  @\newtxt{template<> struct equal_to<void>;}@
+  @\newtxt{template<> struct not_equal_to<void>;}@
+  @\newtxt{template<> struct greater<void>;}@
+  @\newtxt{template<> struct less<void>;}@
+  @\newtxt{template<> struct greater_equal<void>;}@
+  @\newtxt{template<> struct less_equal<void>;}@
```
CaseyCarter: Why are the `foo<void>` specializations separate instead of being declared together with the other `foo` specializations?

CaseyCarter: I guess the obvious answer is "because the detailed descriptions are organized that way, and the descriptions are that way because that's what the Standard does," but I've always found that fact odd as well.

utilities.tex
```diff
@@ -563,7 +596,37 @@
...
+to a built-in operator comparing pointers.
+}
+
+\newtxt{
+\pnum
+For all object types \tcode{T} for which there exists a specialization \tcode{less<T>},
+the instantiation \tcode{less<T>} shall satisfy \tcode{StrictWeakOrder<less<T>, T>()}~(\ref{concepts.lib.callables.strictweakorder}).
+In addition, \tcode{Relation<greater<T>, T>()}~(\ref{concepts.lib.callables.relation}), \tcode{Relation<greater_equal<T>, T>()},
+and \tcode{Relation<less_equal<T>, T>()} shall be satisfied and those class template
+instantiations shall be semantically consistent with \tcode{less<T>}~(\cxxref{operators}).
```
CaseyCarter: `greater` is a `StrictWeakOrder` as well. We need to detail what "semantically consistent" means in this particular case instead of trying to define "semantically consistent": I fear the phrase is too hot and will burn us much like "semantically equivalent" did. I would rather require the specializations to be consistent only if they are valid, so that a user can e.g. define only `less<Widget>` if `greater<Widget>` et al are ill-formed.

utilities.tex
```diff
@@ -563,7 +596,37 @@
 For templates \tcode{greater}, \tcode{less}, \tcode{greater_equal}, and
 \tcode{less_equal}, the specializations for any pointer type yield a total order,
 even if the built-in operators \tcode{<}, \tcode{>}, \tcode{<=}, \tcode{>=}
-do not.
+do not. \ednote{The following sentence is taken from the proposed resolution of
+\href{https://cplusplus.github.io/LWG/lwg-active.html\#2450}{LWG \#2450}.}
+\newtxt{For template specializations \tcode{greater<void>}, \tcode{less<void>},
+\tcode{greater_equal<void>}, and \tcode{less_equal<void>}, if the call operator calls a
+built-in operator comparing pointers, the call operator yields a total order.}
```
CaseyCarter: The ordering induced by the pointer specializations should be required to respect the ordering induced by the built-in operators on pointers. We should draw attention to that requirement with a conspicuous ednote so that the committee can quickly and easily strike it out if there's a good reason the Standard doesn't require it already.

utilities.tex
```diff
@@ -563,7 +596,37 @@
...
+}
+
+\newtxt{
+\pnum
+For all object types \tcode{T} for which there exists a specialization \tcode{equal_to<T>},
+the instantiation \tcode{equal_to<T>} shall satisfy \tcode{Relation<equal_to<T>, T>()},
+and \tcode{equal_to<T>} shall induce an equivalence relation on its arguments.
+In addition, \tcode{Relation<not_equal_to<T>, T>()} shall be satisfied and that class template
+instantiation shall be semantically consistent with \tcode{equal_to<T>}~(\cxxref{operators}).
+}
```
CaseyCarter: ibid.

utilities.tex
CaseyCarter: Since pointer types are object types, this requirement applies to the pointer specializations. We need an ednote to make that fact obvious so that LWG can take exception sooner rather than later.

utilities.tex
```diff
@@ -563,7 +596,37 @@
...
+do not. \ednote{The following sentence is taken from the proposed resolution of
+\href{https://cplusplus.github.io/LWG/lwg-active.html\#2450}{LWG \#2450}.}
+\newtxt{For template specializations \tcode{greater<void>}, \tcode{less<void>},
+\tcode{greater_equal<void>}, and \tcode{less_equal<void>}, if the call operator calls a
+built-in operator comparing pointers, the call operator yields a total order.}
+
+\newtxt{
+\pnum
+Given a relational operator \tcode{\textit{OP}}, an object \tcode{x} of type \tcode{X},
+and an object \tcode{y} of type \tcode{Y}, let \tcode{\textit{BUILTIN_PTR_CMP}(X, \textit{OP}, Y)}
```
CaseyCarter: "object" won't work since `X`/`Y` may be lvalue reference types, this needs the silly "lvalue if `X` is an lvalue reference type or rvalue otherwise" dance.

CaseyCarter: "let XXX be true" should be "XXX shall be true" or "XXX is true".
`BUILTIN_PTR_CMP` needs to be exposition-only. An implementation whose builtin operators provide a total ordering for pointers - as most do - doesn't actually need to do anything to meet the required semantics.

utilities.tex
ericniebler: Correct. I'm not motivated to change the way the standard does this.

utilities.tex
ericniebler: > Since pointer types are object types, this requirement applies to the pointer specializations. We need an ednote to make that fact obvious so that LWG can take exception sooner rather than later.

I guess I don't understand what you're worried about here. The pointer specializations *are* consistent with `less<T*>`.

utilities.tex
ericniebler: > The ordering induced by the pointer specializations should be required to respect the ordering induced by the built-in operators on pointers.

Why? Wouldn't that require the built-in operator to also induce a total ordering? And if we had that, we wouldn't need the pointer specializations.

utilities.tex
CaseyCarter: N4567 says only "For templates `greater`, `less`, `greater_equal`, and `less_equal`, the specializations for any pointer type yield a total order, even if the built-in operators `<`, `>`, `<=`, `>=` do not." It notably does not say that all of them yield the *same* total order, or even that the ordering imposed by `less<Foo*>` is the same as `less<Bar*>`. I think to be sane it must be the case that e.g. `less<Widget*>{}(w1, w2)` is equal to `less<void*>(w1, w2)`, and that the pointer specializations of the other functions must yield the same total order as `less`.

utilities.tex
CaseyCarter: It requires that whenever `a < b` is defined for pointers `a` and `b`, it must be equal to `less<decltype(a)>{}(a,b) == a < b`. I don't want `less` to mean something different than `<` in their common domains.

utilities.tex
CaseyCarter: It's the same general notion as our requirement for cross-type `StrictTotallyOrdered`: the ordering imposed by `<` on `common_type_t<T, U>` must respect the ordering imposed by `<` on `T` and `<` on `U`. I want `a < b == less<>(a, b)` just as we have `t1 < t2 == C(t1) < C(t2)`. They are technically different relations, but should be consistent since they have the same name.

ericniebler:
OK, please take another look, @CaseyCarter

utilities.tex
ericniebler: I'm uncomfortable with this. It gets into platform-y and core-language-y territory. If you feel strongly about this, I suggest you file a library issue and let LWG deal with it.

ericniebler:
Merged in 0f70344

Issue #131, opened 2015-10-20
SizedIteratorRange semantic requirements


CaseyCarter:
Reading through the wiki notes from the Proposal review today, there's a comment asking whether the parentheses can be removed from the requirement expressions:

> * `(b - a) == n`.
* `(a - b) == -n`.
* `(a - a) == 0`.
* `(b - b) == 0`.

Looking at these I realize the comparisons are not well-defined since the syntactic requirements on the expressions use conversion constraints:

```c++
template <class I, class S>
concept bool SizedIteratorRange() {
  return Sentinel<S, I>() &&
    requires (const I i, const S j) {
      { i - i } -> DifferenceType<I>;
      { j - j } -> DifferenceType<I>;
      { i - j } -> DifferenceType<I>;
      { j - i } -> DifferenceType<I>;
    };
}
```

The result types need to be coerced to `DifferenceType<I>` before they can be compared. This is complicated by the fact that only implicit conversion is required, so that explicit conversion to `DifferenceType<I>` may not be well-formed:

* `[n](DifferenceType<I> x){ return x == n; }(b - a)`.
* `[n](DifferenceType<I> x){ return x == -n; }(a - b)`.
* `[](DifferenceType<I> x){ return x == 0; }(a - a)`.
* `[](DifferenceType<I> x){ return x == 0; }(b - b)`.

Which is horrible, but could be somewhat simplified using `ConvertibleTo` once we properly change `ConvertibleTo` to require both implicit and explicit conversions with equal results:

```c++
template <class I, class S>
concept bool SizedIteratorRange() {
  return Sentinel<S, I>() &&
    requires (const I i, const S j) {
      { i - i } -> ConvertibleTo<DifferenceType<I>>;
      { j - j } -> ConvertibleTo<DifferenceType<I>>;
      { i - j } -> ConvertibleTo<DifferenceType<I>>;
      { j - i } -> ConvertibleTo<DifferenceType<I>>;
    };
}
```

so the requirements can become:

* `DifferenceType<I>(b - a) == n`
* `DifferenceType<I>(a - b) == -n`
* `DifferenceType<I>(a - a) == 0`
* `DifferenceType<I>(b - b) == 0`

Issue #107, opened 2015-09-09
Unimplementable requirements for advance and distance


CaseyCarter:
Describing:
```c++
template <Iterator I, Sentinel<I> S>
void advance(I& i, S bound);
```
[iterator.operations]/7 says:
> If `SizedIteratorRange<I,S>()` is satisfied, this function shall dispatch to `advance(i, bound - i)`.

Paragraph 10 similarly requires of:
```c++
template <Iterator I, Sentinel<I> S>
DifferenceType<I> advance(I& i, DifferenceType<I> n, S bound);
```
> If `SizedIteratorRange<I,S>()` is satisfied: ... Otherwise, ...

Paragraph 12 provides a similar requirement for
```c++
template <Iterator I, Sentinel<I> S>
DifferenceType<I> distance(I first, S last);
```
> If `SizedIteratorRange<I,S>()` is satisfied, returns `(last - first)`; otherwise, ...

Since it is not possible for a library implementation to determine if a set of parameters satisfies the non-syntactic requirements of a concept, these requirements are not implementable.

CaseyCarter:
Possible solutions:
* Introduce additional overloads that require `SizedIteratorRange`:
  ```c++
SizedIteratorRange{I, S}
void advance(I& i, S bound);

SizedIteratorRange{I, S}
DifferenceType<I> advance(I& i, DifferenceType<I> n, S bound);

SizedIteratorRange{I, S}
DifferenceType<I> distance(I first, S last);
```
  and similarly for the overloads of `next` and `prev` that are specified to dispatch to `advance`, and any algorithms that may use `advance` or `distance` in their implementations.

* Prohibit `Iterator` / `Sentinel` pairs that meet the syntactic requirements of `SizedIteratorRange` but do not satisfy `SizedIteratorRange`.

ericniebler:
> I don't think the approach is tenable for the algorithms.

I think the correct approach here is to have some text in the library introduction describing what happens when instantiating a template on a type that does not satisfy the requirements: undefined behavior. We would also need to extend this blanket wording to function templates that are specified to use concepts internally to dispatch to optimal implementations.

CaseyCarter:
[...]

PR #153, opened 2015-11-18
Sharpen semantics of Iterators and Sentinels:

CaseyCarter:
* Specify the required definition space of comparison operations on Iterators and Sentinels.
* Rename SizedIteratorRange to SizedSentinel, factor out SizedIterator. Locate both concepts immediately after Iterator and Sentinel, respectively. (The section [iteratorranges] no longer exists.)
* CommonSizedIteratorRange has *not* been renamed.
* Rename disable_sized_range to disable_sized, make it a variable template, and relocate to [iterators.sizediterator]. disable_sized serves as an opt out trait for both SizedIterator and SizedRange (and indirectly SizedSentinel, since it requires its iterator type to satisfy SizedIterator).
* Respecify distance and advance in light of the updated iterator/sentinel/sized semantics.

Closes caseycarter/stl2#13.
Closes caseycarter/stl2#20.
Touches on ericneibler/stl2#88.
Closes ericniebler/stl2#107.
Closes ericniebler/stl2#131.

My approach for `distance` and `advance` was to specify a distinct overload for every syntactically distinguishable precondition. Yes, that results in a dozen or so total overloads of `advance` - not "counting" the overload for `counted_iterator` - which is verging on the ridiculous. I think that more overloads with simpler specifications of requirements and effects result in a more readable spec than would fewer overloads with "Requires: A or B or C"..."Effects: If A then X, otherwise if B then Y, otherwise if C then Z". Opinions will of course vary.


iterators.tex
```diff
@@ -73,7 +73,7 @@
 of the iterator.
 
 \pnum
-Since iterators are an abstraction of pointers, their semantics is
+Since iterators are an abstraction of pointers, their semantics \oldoldtxt{is}\newtxt{are}
```
ericniebler: Nice catch.

iterators.tex
```diff
@@ -623,31 +650,186 @@
...
+\rSec2[iterators.sizediterator]{Concept SizedIterator}
+\pnum
+The \tcode{SizedIterator} concept refines \tcode{Iterator} with the
+capability to determine the distance between two iterators that denote
+elements of the same range in constant time with the \tcode{-} operator.
+
+\indexlibrary{\idxcode{SizedIterator}}%
+\begin{itemdecl}
+  template <class>
+  constexpr bool disable_sized = false;
```
ericniebler: IMO it's a problem that `disable_sized<I>` and `disable_sized<const I>` will yield different results if the user has specialized `disable_sized` on `I`. I'm also uncomfortable that the sentinel type isn't considered when using `disable_sized` to check the `SizedSentinel` concept.

iterators.tex
```diff
@@ -623,31 +650,186 @@
...
+\begin{itemize}
+\item \tcode{i == s}, \tcode{s == i},\tcode{i != s}, and
+      \tcode{s != i} are well-defined.
+
+\item \tcode{bool(i == s)} if and only if \tcode{i} and
+      \tcode{s} denote the same element.
+
+\item If \range{i}{s} and \range{i}{t} denote the same range,
+      then \tcode{s == t} is well-defined and \tcode{s}
+      equals \tcode{t}.
```
CaseyCarter: This requirement breaks a large class of stateful sentinel types. It needs to be removed, and the rest of the changes audited for any dependencies.

ericniebler: I disagree. It follows from the requirement that for two iterators `i` and `j` that are equal, they point to the same element. If `i` equals two sentinels `s` and `t`, then `s` and `t` denote the same element, and they had better compare equal. Sentinels are an abstraction of a past-the-end iterator, and this property had better hold.

iterators.tex
CaseyCarter: The concepts `uncvref` the type before checking `disable_sized`. Explicitly in `SizedRange`, and implicitly in `SizedIterator` in that the library functions take iterators by value so the type comes pre-decayed. I should make that fact explicit in `SizedIterator` as well for clarity. I find it unlikely there's an iterator/range type out there whose size-ability depends on cv-qualification; it's not worth the trouble to try to support such a creature.

Regarding the sentinel type and `disable_sized`: sentinels are new. We don't need to support broken underconstrained sentinels as we do iterators since there are none in the wild.

ericniebler: I'm imagining an iterator that is sized when used with one sentinel but unsized when used with another. What should `disable_sized<My_counted_iterator<int*>>` return? If I use `My_counted_iterator` with `My_counted_sentinel`, the result is sized. If I use it with `unreachable` it isn't.

ericniebler: Oh, you're saying that it's the user's responsibility to property constrain `op-` for sentinels. Hrm. I'm still rather uncomfortable.

iterators.tex
CaseyCarter: Take a gander at this program: https://gist.github.com/CaseyCarter/784067122e7d7de4b665. When `FOO` is 0, two sentinels that denote the end of the same range are not equal. When `FOO` is 1, two sentinels that *compare* equal aren't in fact equal.

The only way sentinels can have this property is if we eliminate a large class of seemingly useful stateful sentinels (that I haven't yet managed to completely characterize). If they're `Regular`, they can only compare equal when they have equal state / are substitutable / are actually equal. If they must be equal when [i,s) and [i,t) denote the same range it greatly limits what exactly the sentinels can do with that state.

iterators.tex
CaseyCarter: Exactly. `unreachable` doesn't have a `-` operator for `My_counted_sentinel`, and user would be a moron to define one if he does not want it.

iterators.tex
ericniebler: You say, "We can't require [i,s) equals [i,t) to imply s == t," but it's equally true to say, "We can't *not* require [i,s) equals [i,t) to imply s == t." It's a problem I grappled with early on in the design. The trouble comes from the sentinel's role as a *logical* (not physical) position in sequence that only makes sense in relation to the iterator it is paired with. It doesn't even make sense to ask two sentinels whether they represent the same position. In relation to what? I suppose the implication here is that sentinels are not `EqualityComparable`, but then iterator/sentinel pairs are *also* not `EqualityComparable`, and the whole thing comes crashing down.

ericniebler: Here is how I've been thinking about this: You can only compare iterators when they refer into the same range, right? That's the domain of `op==` for iterators. Sentinels also have a domain for `op==`. It's only well-defined when they denote the end of the same range. In the `FOO` case where you are doing `assert(s1 == s2);`, you are comparing "singular" sentinels and getting nonsense. I think this is the cleanest way out of the mess, and is consistent with how sentinels are actually used in the library.

iterators.tex
ericniebler: It just feels like a hack to me. I prefer two opt-out hooks: `disable_sized_range<Rng>` and `disable_sized_sentinel<S, I>`. And I don't understand the need for `SizedIterator`. Can't we make do with just `SizedSentinel`?

iterators.tex
```diff
@@ -623,31 +650,186 @@
...
+\end{itemize}
+
+\pnum
+\enternote Particular pairs of types that satisfy
+\tcode{Sentinel} may provide stronger guarantees, but
+\tcode{Sentinel} requires neither that sentinels denoting
+the past-the-end element of distinct ranges be comparable,
+nor that a sentinel and iterator that do not denote a range
+be comparable.
```
ericniebler: This should be normative, and it should be expanded to talk generally about the domain of `op==` for sentinels.

iterators.tex
CaseyCarter: I can compare two `int*` for equality whether or not they refer into the same range. I'm not constrained to the domain that `Iterator` requires to be valid because I know these particular models of `Iterator` are in fact `int*`s. I can similarly compare `s1` and `s2` because they are `S`s and I know that `==` is well-defined for their values - heck, `return true` is well-defined for *all* values. If the semantics of `Sentinel` are in conflict with the semantics of `Regular`, I don't think that `Regular` is the problem. We can't require equality to mean something different when an `S` is an `S` than it does when an `S` is a `Sentinel`.

Transitivity of `==` between iterators and sentinels is the crux of the issue: if `i == s && j == s` implies `i == j`, then a sentinel must represent a position just like an iterator does.

ericniebler: > Transitivity of `==` between iterators and sentinels is the crux of the issue: if `i == s && j == s` implies `i == j`, then a sentinel must represent a position just like an iterator does.

That's exactly my point. `op==` for sentinels has to be consistent with the fact that sentinels represent *position in sequence*. It follows that comparing two sentinels can only make sense when they are denoting (logical) positions in the same sequence. Imagine this case:

```c++
template<Iterator I> struct counted_iterator {
  I it_;
  int count_;
  counted_iterator& operator++() {
    ++it_; ++count_; return *this;
  }
};

struct counted_sentinel {
  int count_;
};

bool operator==(counted_iterator<auto> i, counted_sentinel s) {
  return i.count_ == s.count_;
}

bool operator==(counted_sentinel s1, counted_sentinel s2) {
  return s1.count_ == s2.count_; // BUGBUG right or wrong?
}
```

Now imagine I do this:

```c++
int data[] = {1,2,3,4,5,6,7,8,9};
counted_iterator<int*> it0 { data + 1, 0 }, i1 { data + 1, 1 };
counted_sentinel s0 { 0 }, s1 { 1 };
```

First question: does `it0` equal `it1`? They have the same type and denote the same element ... but not in the same range. Next question: does `s0` equal `s1`? They have different state, but *could* denote the same element, depending on which iterator it's compared to.

It's easy to rearrange the example so that `s0` and `s1` have the *same* state but denote different elements depending on which iterator it's compared to.

IMO, the only logical conclusion is to say that the domain of `op==` for iterators and sentinels has to be limited to comparisons that happen within the same range. `op==` on sentinels has to respect it's semantics: it's a logical position in a range. Its `op==` is free to assume that -- since the user is comparing sentinels -- they must refer to the same range; otherwise, it's a domain violation. And since sentinels always refer to the last-plus-one element, it's always safe for sentinel `op==` to return true, ignoring the state of the sentinel.

iterators.tex
CaseyCarter: `SizedIterator` is distinct for two reasons: (1) `disable_sized`, which obviously is not significant if `disable_sized` becomes `disable_sized_sentinel`, and (2) it makes it slightly more intuitive that functions accepting a `SizedIterator` pair have different preconditions than functions that accept a `SizedSentinel` pair, i.e., "`[i,s)` shall denote a range" vs. "either `[i,j)` or `[j,i)` shall denote a range." I also think the presentation is clearer when iterator requirements are named `FooIterator`; having `RandomAccessIterator` require `SizedSentinel` will throw people off, just as you commented on the issue.

What's the use case for `disable_sized_sentinel`? A user who read the requirements, implemented an improperly constrained `-` anyway for his iterator with any random sentinel, and finds it easier to specialize `disable_sized_sentinel<I, S>` for each problem `S` as it comes up rather than fix his `-`?

iterators.tex
CaseyCarter: This is clarifying the absence of a requirement, it need not be normative. The normative requirement is above in the bullet we've been arguing about.

iterators.tex
ericniebler: Regarding `SizedIterator`, the committee has expressed reservations at the sheer number of concepts. I'd rather have fewer than more. I see `SizedIterator` as one we can do without.

> What's the use case for `disable_sized_sentinel`?

I'm concerned about an unconstrained and unrelated `op-` getting found by ADL. ADL has trained me to be paranoid. Besides, there is the symmetry argument. There are two types involved in the expression `s-i`. The `disable` trait should take two parameters.

iterators.tex
CaseyCarter: Equality in our semantics means substitutable in equality-preserving expressions. `EqualityComparable<T>()` requires that `==` means equal. If `==` for `counted_sentinel` means something else, then it doesn't satisfy `EqualityComparable<counted_sentinel>()`, let alone `EqualityComparable<counted_iterator, counted_sentinel>()` or `Sentinel<...>()`. I think that fixing this problem by changing the semantics of equality is a horrible idea, and that we'd be better served fixing the problem by not requiring iterators and sentinels to satisfy `EqualityComparable`. I don't believe that all sentinels denote positions, so that positions can be the common type of sentinels and iterators; it just doesn't work.

ericniebler: `==` for `counted_sentinel` *does* mean equal. What you don't seem to be grasping is that equality for sentinels is positional, since sentinels denote position. If you blindly compare sentinel state without thinking about what that state represents and how it's used in the algorithms, you're not defining equality for the type. You're defining something else.

iterators.tex
CaseyCarter: I think I can buy the ADL argument, although in that case one could `delete` a more-specialized `op-` overload as a countermeasure. I don't buy the symmetry argument as much; it should take two types to opt in and enable a behavior, but any one is enough to opt out. In any case I don't see this change as a huge detriment. I'll merge two concepts into one and split one trait into two.

iterators.tex
CaseyCarter: It's not possible for two values to be equal, and yet have only one if them be in the domain of an operation. If `i * s1` is an equality-preserving expression that is well-defined for `i` and `s`, and `s1` equals `s2`, then `i * s2` must also be well-defined and equal to `i * s1`. This is the basis for our semantics; it must hold whether or not `i` is an iterator, `*` is `==`, and `s1` and `s2` are sentinels.

So an order for my program to have defined behavior, I have to lie in my `op==`. I then need to ensure that I never pass two instances of my type that are not actually equal to a library component which could then potentially detect that lie. I also need to disable any semantic optimizations in my compiler which may otherwise decide that it need not bother to instantiate `S{foo}` when it already has `s{bar}` lying around in a register because they are equal. It seems like we're injecting a veritable minefield of undefined behavior into the design to support the `EqualityComparable` requirement for sentinels which nothing in the design actually uses or wants.

ericniebler: > If `i * s1` is an equality-preserving expression that is well-defined for `i` and `s`, and `s1` equals `s2`

Stop there. What is `s2`? A sentinel used to denote the end of ... what? Does it even make sense to compare `s1` and `s2`? If they denote the end of the same sequence, then they are in the domain of `==` and will compare equal. If they denote the end of different ranges, then `s1` and `s2` are out of the domain of `==` and it's an error to rely on the result of the operation.

CaseyCarter: It does not matter what `s2` is - or `i` or `*` for that matter - only that it's equal to `s1` and hence has the same type. This argument applies to all values of all types and all equality preserving expressions, because it's what we require in [concepts.lib.general.equality]. If your argument is that it should not matter because generic code should never be permitted to compare sentinels, I'm right there with you - I'd like to enforce that in the semantics by removing the operation. If your argument is that it does not matter because the equality of two different sentinel values is unobservable, then I will point out that is definitely not true. I can look at the program and see that `==` always returns true, so can a compiler. If a program that compares two sentinel values with `==` always has undefined behavior, either because two different values aren't in the domain of `==` or because two "equal" values are in the domain of `==` but they may have observably different behavior, then why have `==` at all?

CaseyCarter: Correction: removing `==` from sentinels would address the issue, but I think a better solution would be to leave sentinels `Regular` and lose the notion that they denote positions and share a common type with iterators. A sentinel is fundamentally a predicate over iterator values: `i == s1` does not imply that `i` and `s1` denote the same position, only that `i` satisfies `s1`'s predicate. Therefore `i == s1 && j == s1` does not transitively imply `i == j` unless `s1` is in fact an iterator.

`s1 == s2` should imply `(i == s1) == (i == s2)` and hence `[i, s1) == [i, s2)`, but it should not imply that `[i, s1)` and `[j, s1)` share a common end position unless `i` is in `[j,s1)` or `j` is in `[i,s1)`.

iterators.tex
CaseyCarter: I'm inclined to require `!disable_sized_sentinel<I, I> && !disable_sized_sentinel<S, I>` for `SizedSentinel` so it's possible to disable sizing for a broken iterator with all possible sentinels in one shot. Speak up if you disagree.

iterators.tex
ericniebler: > EDIT: I suppose it's possible that you're suggesting two equal values may not be in the domain of ==, which would invalidate my argument, but leave me just as baffled.

This is what I've been arguing. Undoubtedly you will get *something* if you compare iterators from different vectors, but you'd be crazy to to anything with the result. Consider also the Range TS's `counted_iterator`, which only compares the internal count. Two counted iterators into different ranges may well compare equal (if their counts are the same), but that doesn't mean they point to the same element! I don't have to tell you that -- you know it already. Sentinels are no different, IMO.

As another example, think of default-constructed `istream_iterator<string>`s (the current STL's equivalent of a sentinel). They all compare equal. That is consistent with their usage: to denote the end of a range. It would be wrong-headed to try to also compare the state of the internal `string` data member!

Given some range `[first,last)` where `last` is a sentinel, and two iterators `i` and `j` that are reachable from `first`, then `i==last && j==last` implies `i==j`. *That* is the essential property of sentinels. And those semantics come straight from the algorithms and how they use iterators, sentinels, and `==`.

So I don't see a need to give up the notion that a sentinel denotes a position or that `==` means equality. `==` over sentinels means equality *in the sense in which `==` and sentinels are used by the algorithms*.

CaseyCarter: I think the disconnect here is that we're arguing opposite meanings. You keep saying that `==` being true doesn't imply two values are equal when those values are not in the domain of `==`, which I agree with unreservedly. What I take issue with is the notion that it's reasonable for two values that *are* in the domain of `==`, for which `==` returns true, to have observably different behavior in equality-preserving expressions. That would contradict the basis of our semantics.

The axiom that began this entire thread states:
```tex
If \range{i}{s} and \range{i}{t} denote the same range,
then \tcode{s == t} is well-defined and \tcode{s}
equals \tcode{t}.
```
This does not imply, but states directly that if `[i,s)` and `[i,t)` denote the same range then `s` and `t` cannot have observably different behavior in equality-preserving expressions. The direct consequence is that for any `j` such that `[j,s)` denotes a range, `[j,t)` must also denote the same range. If `j` happens to be the iterator before `i`, and `t` is satisfied by `j`, we have a contradiction. We cannot say that sentinels only have meaning in connection with a particular range when this axiom gives them meaning across ranges. The axiom effectively requires that the predicates of `s` and `t` both be well-defined and have the same value for every iterator with which either `s` or `t` denotes a range, if there is a single instance in the program of two identical ranges `[i,s)` and `[i,t)`.

I'm sure the softness of the terms "denotes a range" and "same range" are contributing to the cross-talk here. We need to nail them down.

> Given some range [first,last) where last is a sentinel, and two iterators i and j that are reachable from first, then i==last && j==last implies i==j. That is the essential property of sentinels. And those semantics come straight from the algorithms and how they use iterators, sentinels, and ==.

I agree that is the essential property of sentinels; I don't agree with the implication that we need this axiom to ensure that property holds. If anything, it requires "If `[first,last)` denotes a range either `first == last` or `[++first,last)` denotes a range" which is [roughly the definition I have for "denotes a range" in my axioms for sentinels-as-predicates](https://github.com/CaseyCarter/cmcstl2/compare/advance...sentinel_predicates#diff-a54d540f5abe266043bcfee21eac7e99R552).

It's a contradiction in terms for a stateful type whose `==` doesn't reflect that state to satisfy `EqualityComparable`. Either the state doesn't contribute to behavior in EP expressions, in which case it isn't really a stateful type, or the state does have effects observable to EP expressions and the type does not satisfy `EqualityComparable`.

ericniebler: > It's a contradiction in terms for a stateful type whose `==` doesn't reflect that state to satisfy `EqualityComparable`. Either the state doesn't contribute to behavior in EP expressions, in which case it isn't really a stateful type, or the state does have effects observable to EP expressions and the type does not satisfy `EqualityComparable`.

I wholeheartedly agree with this. However, when we talk about EP expressions, we mean *well-defined* EP expressions. Two objects of an `EqualityComparable` type that compare equal may have observably different behaviors in expressions that are ill-formed. If you violate an operation's domain, strange things happen.

I haven't yet tried to specify precisely the domain of `==` for sentinels. I haven't bothered since `==` on sentinels is a pretty useless operation, but you've convinced me that I need to look closer.

CaseyCarter: > Two objects of an `EqualityComparable` type that compare equal may have observably different behaviors in expressions that are ill-formed.

Yes, we're on the same page now. "are equal" is the pre-condition for substitution into expressions, not "compare equal". `EqualityComparable` tells us that "compare equal" and "are equal" mean the same thing if `==` is well-defined for two values. This axiom says these two sentinels are in the domain of `==` and are equal. The result is many bad things.

Please investigate - I think you likely have figured out by now that my position is that the semantics are best served if the domain of `==` for sentinels is the empty set ;)

CaseyCarter: > the semantics are best served if the domain of `==` for sentinels is the empty set

This is an overstatement; making sentinels Semiregular was an easy fix that made the rest of the iterator semantics fall into place nicely, and it also makes weak ranges simple. I haven't yet convinced myself that there's no way for Regular sentinels to work, only that they won't work with this particular axiom.

asutton: > I haven't yet tried to specify precisely the domain of == for sentinels.
I haven't bothered since == on sentinels is a pretty useless operation, but
you've convinced me that I need to look closer.

> + item \tcode{bool(i == s)} if and only if \tcode{i} and
> +      \tcode{s} denote the same element.

Just reading through a recent draft. Sentinels are defined as abstractions
of PTE iterators, so "denote the same element" seems weird.


> +\item If \range{i}{s} and \range{i}{t} denote the same range,
> +      then \tcode{s == t} is well-defined and \tcode{s}
> +      equals \tcode{t}.

My understanding is that sentinels are essentially predicates. So two
functions (expressions, whatever) are equal iff they map the same inputs to
the same outputs. That's tough to implement in C++ :) But you don't really
need to. Representation is a sufficient approximation (function pointer,
function object type, etc).

But this axiom isn't actually true. s and t could be different functions
whose mappings overlap for that particular sequence of elements.

I suppose you could say that two sentinels s and t are equal if, for all
iterators i, i==s equals i==t. Then it would be the case that if s and t
are the same predicate then [i, s) and [i, t) would be the same ranges.

My 2c.

CaseyCarter: > My understanding is that sentinels are essentially predicates.

Yes, that's the gist of it. Sentinels can be Regular, but they cannot be stateful predicates and simultaneously model cross-type `EqualityComparable` with an Iterator. Cross-type EC requires that whenever one iterator is `==` two sentinels, those sentinels are equal.

This axiom is actually a consequence of EC for finite ranges. I added it as an axiom to enable infinite ranges to satisfy EC since one can't presuppose the existence of an iterator equal to the sentinel in an infinite range.

asutton: >
> Yes, that's the gist of it. Sentinels can be Regular, but they cannot be
> stateful predicates and simultaneously model cross-type EqualityComparable
> with an Iterator. Cross-type EC requires that whenever one iterator is ==
> two sentinels, those sentinels are equal.
>
Ah... Of course that's where this is coming from.

When Marcin and I were first writing the Palo Alto paper, we discussed
whether or not the cross-type concepts should require the independent types
to be EC. What appears in the report is what was discussed during the
meeting, but there were some good reasons for relaxing the requirement.

I would say this: even if EqualityComparable<T,U> is satisfied, and for
some t and u, t==u, there is no guarantee that t can be s substituted for u
in all EP expressions. For example, a call f(t) might be well formed, but
f(u) might not --- or worse, you might find an overload with different
semantics than f(t).

This strikes me as one of those cases where i == s just means that i can be
compared for equality with s, but where that capability should impose no
other requirements on the types of i or s.


> This axiom is actually a consequence of EC for finite ranges. I added it
> as an axiom to enable infinite ranges to satisfy EC since one can't
> presuppose the existence of an iterator equal to the sentinel in an
> infinite range.
>
I don't think this is necessary. Since all finite ranges are expected
(required?) to be EC, you should be able to cover them with a semantic
requirement on the concept as a whole.


All finite ranges satisfy are EqualityComparable. For all ranges r1 and r2
(same type), r1 == r2 is true iff r1 and r2 cover (denote? range over?,
contain?) the same elements. That is, iterating over both ranges produces
the same sequence of objects.

That should be sufficient to impose implementation requirements on the
sentinel without precisely saying how it is achieved.

Andrew

CaseyCarter: > Since all finite ranges are expected (required?) to be EC, you should be able to cover them with a semantic
requirement on the concept as a whole.

Yes, I expressed myself poorly here. I meant that this axiom is not required when ranges are finite; it's derivable from the equality of the two ranges. I included it because I'm getting ahead of myself and thinking about infinite ranges.

ericniebler: Just commenting on this part for now:

> Just reading through a recent draft. Sentinels are defined as abstractions
> of PTE iterators, so "denote the same element" seems weird.

A sentinel denotes an element when it compares equal to an iterator, and the iterator points to that element.

ericniebler: > even if EqualityComparable<T,U> is satisfied, and for
> some t and u, t==u, there is no guarantee that t can be s substituted for u
> in all EP expressions.

Here, you seem to be saying that we should loosen the semantic requirements of the cross-type EQ concept to not require substitutability in EP expressions. Which makes sense, but it makes me wonder what cross-type EqualityComparable actually means. Perhaps the meaning is derived from the presence and behavior of the common type in EP expressions.

> This strikes me as one of those cases where i == s just means that i can be
> compared for equality with s, but where that capability should impose no
> other requirements on the types of i or s.

And here you seem to be saying that an iterator and a sentinel should not be required to satisfy EqualityComparable at all. Do you mean both?

CaseyCarter: > Here, you seem to be saying that we should loosen the semantic requirements of the cross-type EQ concept to not require substitutability in EP expressions. 

I believe the point Andrew is making here is that cross-type EC does not guarantee substitutability of `t` and `u` in all EP expressions because it does not require all expressions that are valid for either `t` or `u` must be valid for *both* `t` and `u`, which would be equivalent to requiring the two types to be isomorphic. What cross-type EC *does* guarantee is substitutability of `C{t}` and `C{u}` in all valid EP expressions for the common type `C`.

ericniebler: That was my take also, but I need to sit down with the current draft to make sure it actually says that.

CaseyCarter: > I need to sit down with the current draft to make sure it actually says that.

Abbreviating `EqualityComparable` as EC, and [concepts.lib.compare.equalitycomparable]/x as EC/x:

Let `t` and `u` be objects of types `T` and `U` such that `EC<T, U>()` is satisfied, and  `bool(t == u)` holds. Let `C` be `common_type_t<T, U>`. By EC/3.1, `bool(C{t} == C{u})` must also hold. `EC<T, U>()` requires `EC<C>()` to be satisfied, so by EC/1.2 `bool(C{t} == C{u})` implies that `C{t}` equals `C{u}`. From the definition of equality preservation, `f(C{t})` equals `f(C{u})` for any EP expression `f`.

ericniebler: The problem as I'm understanding it is that if:

```c++
i == s1 && i == s2
```

and if `Sentinel` requires `EqualityComparable`, then then `s1` and `s2` must *be* equal, i.e., substitutable in all equality-preserving expressions. I agree this is a problem. Patching this up in the system as I've been thinking about it would require saying something like, "for a sentinel `s`, the expression `s` is not equality preserving except in the following contexts:....," followed by a list that includes copying, moving, and equality comparison with iterators and other sentinels. In addition, `s1==s2` is defined IIF `[i,s1)` denotes the same range as `[i,s2)` for all `i`. That's getting pretty awful, and I'm ready to entertain alternatives.

CaseyCarter: Yes, essentially. It's very easy with stateful sentinels to construct a range [i,...j,...s1) such that [j,s2) is the same range as [j,s1) but [i,s2) is NOT the same range as [i,s1).

I would begin by removing the requirement that sentinels and iterators are EqualityComparable, and then build up semantics for `i == s` that provide the properties algorithms need to work. We need:
* `i == s` to be equality-preserving (EP).
* `i != s` to be EP and the complement of `==`, i.e., `i == s` iff `!(i != s)`
* `s == i` iff `i == s`
* `s != i` iff `i != s`
* `i == s`, `i != s`, `s == i`, and `s != i` all have the same definition space.
* `i == s` to be well-defined if [i,s) denotes a range.

All but the last are semantics that EC should have and can be presented as a simple relaxation for EC to refine. There's nothing here about `==` for sentinels being valid, or what its domain should be, because the algorithms simply don't care.

asutton: I think this sounds like the right approach.

ericniebler: Agreed.

CaseyCarter:
Fresh pull, with what organically grew into a fairly large change. Start reviewing at Annex F, which is a rationale for the change. In the fewest words possible: we have sentinels, so we don't need distinct weak & strong iterator categories. I've implemented the proposed change on branches of both cmcstl2 and range-v3, so it's not completely insane.

sentrework.tex
```diff
@@ -0,0 +1,230 @@
...
+
+\pnum
+The first three properties aren't specific to iterators and sentinels; they are
+general characteristics of the operators \tcode{==} and \tcode{!=} that are
+intuitively expected to hold whenever they are defined. As such, the
+\tcode{EqualityComparable} concepts should require them to hold and this
+particular set of properties can then be presented as a relaxation of
+\tcode{EqualityComparable<T,U>()}. The current design does exactly that,
+defining the concept
+\tcode{Bike\-shed\-Equal\-ity\-Comp\-arable}~(\ref{concepts.lib.compare.equalitycomparable})
```
ericniebler: Let's suggest a plausible name for this concept. There's a chance that the Ranges TS could be approved at the next meeting, and we want to be ready for that.

sentrework.tex
```diff
@@ -0,0 +1,230 @@
...
+thereof.
+
+\pnum
+Notably, nothing in this list of properties requires comparison of sentinels
+with other sentinels -- it's not a useful feature for generic code. The
+requirement for equality comparison of sentinels in N4560 is a checkbox feature
+to enable satisfaction of cross-type \tcode{EqualityComparable}. Absent the
+\tcode{EqualityComparable} requirement, nothing in the design requires equality
+comparison for sentinels. It's nonsensical to require sentinels to have
+\tcode{Regular} types if \tcode{==} is never required to be valid or useful,
```
ericniebler: "...or useful; they ..." <-- semicolons are not just for C++.

sentrework.tex
```diff
@@ -0,0 +1,230 @@
...
+iterator types -- that the algorithms do not admit. It would appear that
+algorithms that accept input (output) ranges should possibly be relaxed to
+accept weak input (weak output) ranges instead. How would doing so constrain
+the implementations of the algorithms? Do weak ranges have less expressive power?
+
+\rSec2[sentrework.ram.singlepass]{Single-pass Algorithms}
+
+\pnum
+Algorithms that accept input or output iterators necessarily perform at most one
+pass over the range of elements those iterators reference, N4560's weak input and
```
ericniebler: "...iterators reference. N4560's ..."

sentrework.tex
```diff
@@ -0,0 +1,230 @@
...
+weak output iterators are consistent with those single-pass semantics. The
+distinction is obviously support for equality comparison. How do algorithms use
```
ericniebler: Incomplete sentence.

sentrework.tex
```diff
@@ -0,0 +1,230 @@
...
+pointless to compare the iterator with itself or a copy; there's no reason to
+test things known to be equal for equality. Consequently, relaxing the
+algorithms that use single-pass ranges to ``weak'' single-pass ranges doesn't
+reduce their power.
+
+\rSec2[sentrework.ram.merge]{Why ``strong'' Iterators?}
+
+\pnum
+Relaxing the algorithms necessitates relaxing the iterator operations
+(\tcode{advance} et al~(\ref{iterator.operations})) which the algorithms use.
```
ericniebler: "\tcode{advance} et al.~(...), which"

sentrework.tex
```diff
@@ -0,0 +1,230 @@
...
+Indeed there are many components in the specification that require a ``strong''
+iterator not because they need to compare iterator values for equality, but
+because iterator / sentinel pairs in N4560 necessarily used strong iterators.
+Once all such occurrences are relaxed it becomes apparent that there is no need
+for separate ``weak'' and ``strong'' concepts for input and output iterators,
```
ericniebler: semicolon

sentrework.tex
```diff
@@ -0,0 +1,230 @@
...
+After removing the distinction between weak and strong input and output
+iterators, the \tcode{Iterator} concept is insufficiently distinguished from
+\tcode{Sentinel} to deserve a separate named concept. The current design pushes
+\tcode{Iterator}'s \tcode{Sentinel<I,I>()} requirement up into
+\tcode{ForwardIterator} so that the semantics of forward and stronger iterators
+are unchanged, and pushes the \tcode{EqualityComparable} requirement from
+\tcode{Iterator} down into \tcode{Sentinel}, but only requires it when the
+iterator and sentinel types are the same. \tcode{Iterator} may then be struck
+from the design, and the \tcode{Weak} prefix stripped from \tcode{WeakIterator}
+as well.
```
ericniebler: Even if it is superfluous, I would prefer keeping a concept named "Iterator". It's absence would be too conspicuous.

concepts.tex
```diff
@@ -43,15 +43,21 @@
...
 output of an expression is the expression's result and all operands modified by the expression.
-\enternote Not all input values must be valid for a given expression; e.g., for integers \tcode{a}
+
+\pnum
+Not all input values must be valid for a given expression; e.g., for integers \tcode{a}
 and \tcode{b}, the expression \tcode{a / b} is not well-defined when \tcode{b} is \tcode{0}. This
-does not preclude the expression \tcode{a / b} being equality preserving. \exitnote
+does not preclude the expression \tcode{a / b} being equality preserving. \newtxt{The
+\techterm{definition space} of an expression is the set of input values for which the
+expression is well-defined.}
```
ericniebler: The current standard already uses the term "*the domain of*". Perhaps we should bless that as a Phrase of Power.

concepts.tex
```diff
@@ -527,14 +533,42 @@
 \ednote{Remove table [equalitycomparable] in [utility.arg.requirements]. Replace references to
 [equalitycomparable] with [concepts.lib.compare.equalitycomparable].}
 
+{\color{newclr}
+\indexlibrary{\idxcode{BikeshedEqualityComparable}}%
```
ericniebler: I suggest "`WeakEqualityComparable`".

concepts.tex
```diff
@@ -527,14 +533,42 @@
...
+    { t == u } -> Boolean;
+    { u == t } -> Boolean;
+    { t != u } -> Boolean;
+    { u != t } -> Boolean;
+  };
+}
+\end{itemdecl}
+
+\begin{itemdescr}
+Let \tcode{t} and \tcode{u} be objects of types \tcode{T} and \tcode{U}.
```
ericniebler: Virtual Stephan says, "What is 'object' here? What about references? Why not use `declval`?"

concepts.tex
```diff
@@ -542,20 +576,20 @@
 \pnum
 Let \tcode{a} and \tcode{b} be objects
 of type \tcode{T}. \tcode{EqualityComparable<T>()}
-is satisfied if and only if
+is satisfied if and only if:
 
 \begin{itemize}
-\item \tcode{bool(a == a)}.
+\item \oldtxt{\tcode{bool(a == a)}.}
```
ericniebler: Implied by equality preservation, right?

concepts.tex
```diff
@@ -576,16 +611,22 @@
...
+\item \oldtxt{\tcode{bool(b != a) == bool(C(b) != C(a))}.}
 \end{itemize}
 \end{itemdescr}
 
+{\color{newclr}
+\pnum
+\enternote There is no syntactic difference between \tcode{BikeshedEqualityComparable<T, U>()} and
+\tcode{Equal\-ity\-Comp\-arable<\brk{}T, \brk{}U>()}.\exitnote
+} %% color{newclr}
+
```
ericniebler: I puzzled over this for a few seconds. Consider saying this as saying the differences are purely semantic.

intro.tex
```diff
@@ -73,13 +73,13 @@
 
 \begin{itemize}
 \item ISO/IEC 14882:2014, \doccite{Programming Languages - \Cpp}
-\item JTC1/SC22/WG21 N4377, \doccite{Technical Specification - \Cpp Extensions for Concepts}
+\item ISO/IEC TS 19217:2015, \doccite{Programming Languages - \Cpp Extensions for Concepts}
```
ericniebler: Great, thanks.

ericniebler:
Blast github for not showing a diff of iterators.tex.

sentrework.tex
CaseyCarter: "distinction is..support" is complete, the "obviously" is out of place. "The sole distinction is support for equality comparison." is more direct and reads better.

concepts.tex
CaseyCarter: That works for me.

sentrework.tex
CaseyCarter: `WeakIterator` is now `Iterator`, which aligns nicely with the usage of "iterators" in the prose.

concepts.tex
CaseyCarter: Yes; it's in the note.

concepts.tex
CaseyCarter: I think that would be clearer - I'll change the other similar notes as well.

concepts.tex
CaseyCarter: Cut'n'pasted from the other EC concepts. I was considering changing them to "values" instead of "objects," but decided it can wait until I decide how to handle the "value type" concepts uniformly. As currently specified, `int(double, char)` satisfies `EqualityComparable`, and `EqualityComparable<int[42],int[]>()` is satisfied as well. The whole mess is detailed in #155.

sentrework.tex
CaseyCarter: We'll go with "Weak" until something better comes along. Once I started laughing at the double meaning of `UncommonlyEqualityComparable` I was incapable of thinking of a reasonable name.

StephanTLavavej:
You probably need wording somewhere that says, "when we want BLAH1 op BLAH2 to work, we require all constness and value category permutations to work".

ericniebler:
We already have such blanket wording in the library introduction.

CaseyCarter:
> You probably need wording somewhere that says, "when we want BLAH1 op BLAH2 to work, we require all constness and value category permutations to work".

That requirement is in the general discussion on equality-preserving expressions, specifically [concepts.lib.general.equality]/6:

> Where a requires-expression declares an expression that is non-modifying for some constant lvalue operand, additional variants of that expression that accept a non-constant lvalue or (possibly constant) rvalue for the given operand are also required except where such an expression variant is explicitly required with differing semantics. ...

ericniebler:
That's all  the feedback I have for now. Looking good.

CaseyCarter:
There's the commit with corrections for the review comments. Make sure you're happy with everything in it, and I'll squash/rebase/fix the merge conflicts.

ericniebler:
Maybe `Weak`**`ly`**`EqualityComparable` would be better?

CaseyCarter:
> Maybe WeaklyEqualityComparable would be better?

Both are sensible to me: "Weakly" is an adverb that modifies the adjective "EqualityComparable", and "Weak" is an adjective applied to the proper noun "EqualityComparable" that names a concept. I have no strong feelings either way.

ericniebler:
I would be happier with `WeaklyEqualityComparable`. Which makes me think it should be `StrictlyTotallyOrdered`, although it's a mouthful.

ericniebler:
Aside from those small tweaks, this is looking good.

CaseyCarter:
There's a squashed and rebased commit with the `WeakEqualityComparable` => `WeaklyEqualityComparable` change. I think we're done here.

PR #157, opened 2015-12-09
Merge{_move} fixes:

CaseyCarter:
* Require the output iterator to satisfy WeaklyIncrementable instead of Incrementable.
* Specify operational semantics in place of the ill-formed (and possibly meaningless) postcondition.


algorithms.tex
```diff
@@ -879,14 +875,14 @@
     \ednote{A new algorithm, needed by \tcode{inplace_merge} and \tcode{stable_sort}.}
   \begin{codeblock}
   template<InputIterator I1, Sentinel<I1> S1, InputIterator I2, Sentinel<I2> S2,
-      Incrementable O, class Comp = less<>, class Proj1 = identity,
+      @\newtxt{Weakly}@Incrementable O, class Comp = less<>, class Proj1 = identity,
```
ericniebler: Once we fix `[Move]Writable`, we can drop `merge_move`.

Issue #37, opened 2015-05-29
common_iterator needs some help to interface STL2 code to STL1 code.


CaseyCarter:
`common_iterator` is presented as a facility for interfacing "new" code that uses STL2 iterator/sentinel pairs with "old" code that accepts iterator pairs. There are a couple of shortcomings in this design:
1. Converting to `common_iterator` is not always the most efficient way to derive an iterator pair from an iterator/sentinel pair. The fact that `common_iterator` requires the iterator and sentinel types to be different is a nod to the fact that using a `common_iterator` is less efficient than using the original type directly in this case. That requirement unfortunately makes `common_iterator` impossible to use directly in generic code. Additionally, if the iterator models `RandomAccessIterator` and together with the sentinel models `SizedIteratorRange`, it is possible to convert the sentinel to the iterator type in constant time resulting in a more efficient common type than `common_iterator`.
2. Given the differences in the iterator models of STL2 and STL1 - particularly the STL1 requirement that the reference type of a `ForwardIterator` be a true reference which is not the case for STL2 - an STL2 iterator may not model the STL1 iterator category that is identified by the same iterator category tag.

I think we have a duty to ease the STL1 to STL2 transition as much as possible, so it is important to facilitate interfacing "new" iterator code to "old" iterator code for codebases in transition. I propose a design for two facilities to address the above problems. First, a family of function templates `make_common_iterator` that converts an iterator or sentinel in constant time into a common type that models the most refined iterator concept supportable by that iterator and sentinel, and second a family of function templates `decay_iterator` that adapts an STL2 iterator into the most refined STL1 iterator concept that its type models.
### `make_common_iterator` family

```
// make_common_iterator<S>(I) transforms the given InputIterator into a common
// type of S and I that models the most refined iterator concept supportable by
// S and I with no more than constant time conversion complexity.

// Base iterator template
template <class S, InputIterator I>
    requires Sentinel<S, I>
auto make_common_iterator(I i) {
    return common_iterator<I, S>{i};
}

// If S and I are the same type, they are already Common.
template <class S, InputIterator I>
    requires Same<S, I>
auto make_common_iterator(I i) {
    return i;
}

// If S and I model SizedIteratorRange and I models RandomAccessIterator, we
// can convert the sentinel to I in constant time - so we need not convert i.
template <class S, RandomAccessIterator I>
    requires SizedIteratorRange<I, S>
auto make_common_iterator(I i) {
    return i;
}

[...]

PR #160, opened 2015-12-16
Merge Writable and MoveWritable concepts

CaseyCarter:
...into a single concept `Writable<O, T>` which means "an expression whose `decltype` is `T` can be assigned to the result of dereferencing an object of type `O`." This adds some complexity to the design:

* the change to `Writable` cascades to `OutputIterator` and `OutputRange`
* the change to `Writable` necessitates additional requirements for algorithms that want to cache values in temporaries before writing, which I've wrapped into named concepts `LazyIndirectlyCopyable` and `LazyIndirectlyMovable`. (No, I'm not married to those names. They don't matter much given that they currently appear in only two places: `Permutable` and the "no-`ForwardIterator`" overload of `unique_copy`.) `LazyXXX` is exactly the formulation of `XXX` from P0022, but given a distinct name to avoid overconstraining algorithms that don't need temporaries.

The added complexity will be required to integrate P0022's proxy iterators in any case.

The corresponding pro for these cons is that "copy" algorithms once again work with `move_iterator`, allowing the elimination of `MoveMergeable`, `merge_move`, and `partition_move`.

Closes caseycarter/stl2#17.


ericniebler:
My preference, at first brush, is to just make `Indirectly(Mov|Copy)able` stronger and let the other algorithms be overconstrained. If it's necessary to have a weaker concept for those other algorithms, maybe `WeaklyIndirectlyCopyable`. Blech.

iterators.tex
```diff
@@ -342,55 +342,59 @@
...
-Let \tcode{v} be an rvalue or a lvalue of type (possibly \tcode{const}) \tcode{T}, and let \tcode{o}
-be a dereferenceable object of type \tcode{Out}. Then \tcode{MoveWritable<Out, T>()} is satisfied if and only if
+Let \tcode{v} be an \oldtxt{rvalue or a lvalue of type (possibly \tcode{const})}\newtxt{an expression such that \tcode{decltype(v)} is} \tcode{T}, and let \tcode{o}
+be a dereferenceable object of type \tcode{Out}. Then \tcode{\oldtxt{Move}Writable<Out, T>()} is satisfied if and only if
 
 \begin{itemize}
-\item If \tcode{Out} is \tcode{Readable},
-then after the assignment \tcode{*o = std::move(v)}, \tcode{*o} is equal
+\item If \oldtxt{\tcode{Out} is }\tcode{Readable<Out>()\newtxt{ \&\& Same<value_type_t<Out>, decay_t<T>{>}()}}\newtxt{ is satisfied},
+then\newtxt{ \tcode{*o}} after the assignment\oldtxt{ \tcode{*o = std::move(v)}, \tcode{*o}} is equal
```
ericniebler: You can make this stronger by changing `Same` to `Common`.

iterators.tex
```diff
@@ -342,55 +342,59 @@
...
-in those requirements must work as specified whether \tcode{v} has been moved
+\oldtxt{\tcode{v}'s state}\newtxt{If \tcode{v} is an xvalue~(\cxxref{basic.lval}), the resulting
+state of the object it denotes} is unspecified. \enternote \oldtxt{\tcode{v}}\newtxt{The object} must still meet the
+requirements of \oldtxt{the}\newtxt{any} library component that is using it. The operations listed
+in those requirements must work as specified whether \oldtxt{\tcode{v}}\newtxt{the object} has been moved
 from or not.\exitnote
 
+\ednote{``If \tcode{v} is an xvalue'' suffices for now, but incorporating P0022's proxy
+ iterators will require a characterization of the ``xvalue-ness'' of a proxy reference type.}
+
```
ericniebler: Fun.

iterators.tex
CaseyCarter: I considered it. If `Readable<In>() && Common<value_type_t<Out>, decay_t<T>>()` then `C(*o)` equals `C(v)` where `C` is `common_type_t<...stuff...>` would be sensible, but would forbid "lossy" conversions. For example, copying a range of doubles to a range of ints with truncation.

CaseyCarter:
> My preference, at first brush, is to just make Indirectly(Mov|Copy)able stronger and let the other algorithms be overconstrained.

I assumed that would be the case, since that's exactly what you did with P0022. Strengthening `IndirectlyCopyable` is a non-starter: it forbids using "copy" algorithms with move-only types via `move_iterator`, and puts us right back where we started. The case for `IndirectlyMovable` is obviously much weaker; requiring a `Movable` value type is not a huge issue. I think the design is more comprehensible if we treat both cases consistently.

iterators.tex
CaseyCarter: What I would *like* to require is that there must exist a function `f(decay_t<T>)` such that `*o = v` inserts an element into the output sequence with value `f(v)`, but it seems too ridiculously abstract to bother.

CaseyCarter:
> puts us right back where we started.

Not completely - algorithms constrained directly with `Writable`, e.g. `transform`, would still be better off.

iterators.tex
ericniebler: > For example, copying a range of doubles to a range of ints with truncation.

&lt;sigh&gt;

ericniebler:
`partition` requires forward iterators, and we demoted `move_iterator` to input. So we can't remove `partition_move`. I can't escape the feeling that  using `move_iterator` to make `_copy` algorithms perform moves is a hack, and I'm becoming uncomfortable with the design compromises we're making to perpetuate it. Let me think about it.

CaseyCarter:
The "copy" equivalent of `partition_move` is `partition_copy`, which requires input iterators.

ericniebler:
D'oh.

asutton:
>
>
> The authors of N3351 thought move_iterator was useful:
>
> Note that these algorithms [copy, copy_n, copy_if, copy_backward] can be
> instantiated in such a way that the input sequence is invalidated by the
> operation. Passing move_iterators as the input range will cause the
> algorithm to move (not copy!) the values into the output range. This will
> invalidate the original iterators; their referenced objects are left
> partially formed. Because we can do this, and because it is sometimes
> useful to do so, we do not require that copy preserves the original input.
>
>
As an author... :)

Marcin and I discussed this *at length* when writing the paper. Having had
more time to think about this, I agree with Eric that it feels like a hack.
I believe that move iterators subvert the non-modification guarantee
[25.1p4].

   1.

   For purposes of determining the existence of data races, algorithms
   shall not modify objects referenced through an iterator argument unless the
   specification requires such modification.

Maybe it's not a big deal. After all, there are viable uses. But use with
care.

I'd hate to have to specify duals of all the copying algorithms.
>

That might not be a bad thing.

ericniebler:
>> I'd hate to have to specify duals of all the copying algorithms.\
>
> That might not be a bad thing.

Would we really need dupes of *all* the copy algorithms? Most of the copy algorithms only copy *some* of the elements out of a sequence. How often do you want to move only some of the elements out of a sequence, leaving holes that have unspecified state?

CaseyCarter:
> I believe that move iterators subvert the non-modification guarantee [25.1p4].
>>  For purposes of determining the existence of data races, algorithms
   shall not modify objects referenced through an iterator argument unless the
   specification requires such modification.

One could argue that this should be reworded to agree with the intent of [res.on.arguments]/1.3:
> If a function argument binds to an rvalue reference parameter, the implementation may assume that
this parameter is a unique reference to this argument.

By passing an iterator with an xvalue reference type to an algorithm, I give that algorithm license to modify the referenced objects. If it's a "unique reference" to that object, data races aren't possible.

It *would* be nice to be able to say that "copy always copies, and move always moves," but it's not consistent with the core language. We don't have distinct copy-assignment and move-assignment operators, we use `=` for both and the effect that `=` has depends on the cv-qualifiers and value category of the RHS. `std::copy` is essentially a range assignment; it *should* act like `=`.

> Does the world really need `remove_move_if`?
Do we add separate move-construct-from-range and copy-construct-from-range overloads to the containers? Does the world really need `view::move`?

asutton:
>
> I believe that move iterators subvert the non-modification guarantee
> [25.1p4].
>
> For purposes of determining the existence of data races, algorithms
> shall not modify objects referenced through an iterator argument unless the
> specification requires such modification.
>
> One could argue that this should be reworded to agree with the intent of
> [res.on.arguments]/1.3:
>
> If a function argument binds to an rvalue reference parameter, the
> implementation may assume that
> this parameter is a unique reference to this argument.
>
> By passing an iterator with an xvalue reference type to an algorithm, I
> give that algorithm license to modify the referenced objects. If it's a
> "unique reference" to that object, data races aren't possible.
>

I agree with this change in principle, but I don't think it addresses the
same issue.

It could be the case, for example, that an algorithm taking move_iterators
actually moves nothing because no "copy/moves" (for utter lack of a better
word) are performed (i.e., everything referenced by values). It could also
be the case that an algorithm unknowingly invokes a copy/move by calling a
user-provided function argument.

Whether or not the use of a move_iterator happens to be modifying is almost
entirely outside the scope of the algorithm.


> Does the world really need remove_move_if?
>
> Do we add separate move-construct-from-range and copy-construct-from-range
> overloads to the containers? Does the world really need view::move?
>
Heh. Bjarne gave the implementation of this algorithm when he was teaching
at TAMU except that it was called extract_if. And the simplest
implementation was remove_if with move iterators :)

Maybe a good answer for now is to simply vacate non-modification and data
guarantees for the use move_iterators. Basically: "these probably work, but
you're on your own".

Andrew

ericniebler:
I just had a crazy and dangerous thought. What if the move adaptor wrapped the *output* iterator instead of the input iterator? In other words, write a `move_from_iterator` that wraps an iterator `it` and has an  `operator=(auto &t)` that basically does `*it = std::move(t)`. Then the `move` algorithm is just `copy(first, last, move_into(out))`?

This is dangerous because nothing is stopping such an algorithm from trying to re-read a value from the input that has already been moved from. There may be other very good reasons why this is a horrible idea. Just thought I'd get it out there.

> Do we add separate move-construct-from-range and copy-construct-from-range overloads to the containers?

Possibly!

> Does the world really need `view::move`?

Possibly not!

asutton:
>
> I just had a crazy and dangerous thought. What if the move adaptor wrapped
> the *output* iterator instead of the input iterator? In other words,
> write a move_from_iterator that wraps an iterator it and has an operator=(auto
> &t) that basically does *it = std::move(t). Then the move algorithm is
> just copy(first, last, move_into(out))?
>
That's crazy enough that it just might work :)


> This is dangerous because nothing is stopping such an algorithm from
> trying to re-read a value from the input that has already been moved from.
> There may be other very good reasons why this is a horrible idea. Just
> thought I'd get it out there.
>

If you limit the usage to only places where output iterators are used, then
it wouldn't be a problem since you can't read from them anyways. So this
would be perfect for the set_* algorithms, which was one of my favorite use
cases for move iterators to begin with.

Andrew

CaseyCarter:
> What if the move adaptor wrapped the output iterator instead of the input iterator?

This seems to address the same problem as making `move_iterator` input-only - ensuring that each element is moved at most once - but does so without restricting the traversal category of the input range. It would therefor be usable with e.g. `reverse_copy` and `rotate_copy` that currently won't work with `move_iterator`.

We'd need to inject a customization point into `*o = *i` to get it to work with proxy iterators, assuming that making `move` a customization point is still off the table. Something like `iter_copy(i, o)` or `iter_assign(o, i)`, defaulting to `*o = *i`, customized for `move_into_iterator<O>` as `*o = iter_move(i)`.

CaseyCarter:
> assuming that making `move` a customization point is still off the table

Although I'm staring to wonder if a single `proxy_move` customization point could solve all the problems addressed by:
* `iter_move` (`foo = proxy_move(*bar)`),
* `iter_swap` (implement `swap` as `auto tmp = proxy_move(a); a = proxy_move(b); b = move(tmp);`)
* `iter_copy/assign` (`move_into_iterator`'s `operator=(T&& t)` is simply `*wrapped_iterator = proxy_move((T&&)t)`)

ericniebler:
>> This is dangerous because nothing is stopping such an algorithm from trying to re-read a value from the input that has already been moved from.
>
> If you limit the usage to only places where output iterators are used, then
it wouldn't be a problem since you can't read from them anyways.

What I mean is that code like this would be broken:

```c++
*out++ = *i; // Actually, this is a move
*out++ = *i; // Oops, *i has been moved from!
```

It's the old `move_iterator` problem reborn. I'm not sure how serious it is, though.

> Although I'm staring to wonder if a single `proxy_move` customization point could solve all the problems 

That's not terrible. I guess the next step is to implement all this and see how broken it is.

ericniebler:
I'm inclined to postpone merging this pending investigation into the `move_into_iterator` suggestion. Somebody shout if you think otherwise.

CaseyCarter:
The fundamental change in this commit is the change to `Writable` to ensure that `*o = foo` is properly constrained on the actual type and value category of the expressions `foo` that an algorithm wants to assign into its output sequence, instead of simply choosing between `value_type_t<In>&&` with `MoveWritable` and `const value_type_t<In>&` with `Writable`. That change is necessary to properly constrain copying algorithms using either `move_iterator` or `move_into_iterator`, or *any* algorithm with an input iterator whose reference type is not possibly-`const` `value_type_t<In>` possibly-reference (proxy iterators). `transform` also needs a way to constrain its output iterator to accept `indirect_result_of_t<...>` without requiring its type to be copyable.

As currently specified, `IndirectlyCopyable` algorithms on move-only types with `move_into_iterator` will fail for the same reason as with `move_iterator`: they require `*o = declval<const value_type_t<In>&>()` to be a valid expression despite that it is not actually be used.

iterators.tex
```diff
@@ -4287,6 +4394,173 @@
...
+\begin{itemdescr}
+\pnum
+\effects Equivalent to \tcode{i.current - s.last}.
+\end{itemdescr}
+
+\indexlibrary{\idxcode{make_move_sentinel}}%
+\begin{itemdecl}
+template <Semiregular S>
+  move_sentinel<S> make_move_sentinel(S s);
```
ericniebler: This is a potential pessimization. There should be an overload of `make_move_sentinel` that takes an `InputIterator` and returns a `move_iterator`. That way, a bounded `[I,I)` range is not turned into a non-bounded `[move_iterator<I>,move_sentinel<I>)` range in generic contexts.

ericniebler:
I suggest changing `LazyIndirectlyMovable` to `IndirectlyMovableStorable`. It's also not a great name, but it's more suggestive of its purpose. Other than that, I'm OK with this change.

iterators.tex
CaseyCarter: >  I suppose you could argue that since move_iterator is Input, it doesn't matter.

I might argue that it's not a pessimization since `move_iterator` is always input, but there's clearly a loss of information. This is almost the identical discussion we had about common iterators months ago in ericniebler/stl2#37; separately converting an iterator and sentinel is generally less effective than converting ranges to ranges due to lack of information about the relationship between the iterator and sentinel. The argument there for `make_bounded_range` applies here as well: `make_move_range` would be more effective, but encroaches on `view` territory.

I'd prefer to leave this as-is and solve the issue with `view::move` rather than specifying ad-hoc `make_foo_range` functions that will be redundant once views are added.

PR #185, opened 2016-10-06
First changes

cjdb:
I've applied the first few changes from [ranges review jacksonville.txt](https://github.com/ericniebler/stl2/blob/master/review/ranges%20review%20jacksonville.txt), barring

p9
s/Range [i,s)/A range [i,s)/

which appears to have been completed already, and

24.2.3 Writable
p/2, change the name of 'v' in the paragraph

which might be completed?

Next to implement: p/4 of SizedSentinel.


CaseyCarter:
+1 from me - these look good.

ericniebler:
Thanks.

PR #186, opened 2016-10-07
Final changes from 'ranges review jacksonville.txt'.

cjdb:
Final changes from 'ranges review jacksonville.txt', excluding the Writable section (should I report an issue for this one?).

@ericniebler, I think you had your eye on one of the proxy iterator reviews: are there any that I shouldn't pick right now?


ericniebler:
Don't worry about the proxy iterators paper and the review feedback for it. I'm making those edits now.

iterators.tex
```diff
@@ -1069,7 +1070,7 @@
 \rSec2[indirectcallables.as_function_t]{as_function_t type}
 
 \pnum
-The \tcode{\oldtxt{FunctionType}}\tcode{\newtxt{as_function_t}} is an alias used to turn a
+The \tcode{\oldtxt{FunctionType}}\tcode{as_function_t} is an alias used to turn a
```
ericniebler: I can't see on my phone. Is this whole section removed? Is that why `\newtxt` is wrong?

CaseyCarter: This whole section is `oldclr`; `as_function_type` is no more after the replacement of Functions with Callables.

CaseyCarter:
> excluding the Writable section (should I report an issue for this one?)

IIRC, there was confusion because of the reuse of the name `v` to refer to both the argument of the requires clause in the concept definition, and the invented expression in the textual description of the semantic constraints. We can address this most easily by replacing `v` with `t` in the concept definition.

cjdb:
Noted, I'll put that into a new pull request.

Issue #187, opened 2016-10-07
Is IndirectlyMovableStorable underconstrained?


ericniebler:
I want to add the following to the specification of `IndirectlyMovableStorable`:

3. Let `i` be a dereferenceable object of type `In`, let `o` be a 
dereferenceable object of type `Out`. Then `IndirectlyMovable<In, Out>()`
is satisfied if an only if
(3.1) -- If `Out` satisfies `Readable`, the expression `*o` has the same value after either `*o = iter_move(i)` or `*o = value_type_t<In>(iter_move(i))`.

@CaseyCarter Thoughts?

CaseyCarter:
Since `rvalue_reference_t<In>` is the same type as `decltype(iter_move(i))`, `*o = iter_move(i)` is constrained to have those semantics by `IndirectlyMovable<In, Out>()` which requires `Writable<Out, rvalue_reference_t<In>>()`.

`*o = value_type_t<In>(iter_move(i))` is derivable from `IndirectlyMovableStorable<In, Out>()`. It requires both `Writable<Out, value_type_t<In>>()` and `Constructible<value_type_t<In>, rvalue_reference_t<In>>()`.

PR #190, opened 2016-10-08
proxy iterators review feedback, customization point objects, update for N4569

ericniebler:


ericniebler:
@CaseyCarter please give this a look over if you get a chance. In particular, have a look at the new definitions of the `iter_move` and `iter_swap` customization point objects. I decided to do it slightly differently than they way you defined `swap`.

ericniebler:
I also strengthened `IndirectlyMovableStorable` and `IndirectlyCopyableStorable` per #187. This might be overreaching a bit.

ext/proxy/Dxxxx.md
```diff
@@ -877,207 +890,155 @@ Change concept `Readable` ([readable.iterators]) as follows:
...
+> struct projected {
+>   using value_type = decay_t<indirect_result_of_t<Proj&(I)>>;
+>   indirect_result_of_t<Proj&(I)> operator*() const;
+> };
+> ```
+
+
+Remove the definition of the `rvalue_reference_t` type alias in  `IndirectlyMovable` ([commonalgoreq.indirectlymovable]).
+
+Change the description of the `IndirectlyMovableStorable` concept ([commonalgoreq.indirectlymovable]) to be:
```
CaseyCarter: As I commented in #187, this is derivable and therefore need not be an axiom.

ext/proxy/Dxxxx.md
```diff
@@ -877,207 +890,155 @@ Change concept `Readable` ([readable.iterators]) as follows:
...
+> 3\. Let `i` be a dereferenceable object of type `In`,
+> let `o` be a dereferenceable object of type `Out`. Then
+> `IndirectlyMovableStorable<In, Out>()` is satisfied if
+> and only if
+>
+> (3.1) -- If `Out` satisfies `Readable`, the expression `*o`
+> has the same value after either `*o = iter_move(i)` or 
+> `*o = value_type_t<In>(iter_move(i))`.
+
+Change the description of the `IndirectlyCopyableStorable` concept ([commonalgoreq.indirectlycopyable]) to be:
```
CaseyCarter: Same here: both `value_type_t<In> v = *i` and `*o = v` are individually required by the algorithms and must be supported by `IndirectlyCopyableStorable`; the semantics of the composition of the two expressions is derivable.

ext/proxy/Dxxxx.md
```diff
@@ -1135,63 +1089,91 @@ Change 24.6 "Header `<experimental/ranges/iterator>` synopsis" ([iterator.synops
...
-> 2\. The expression in the `noexcept` is equivalent to:
-> > ```c++
-> > noexcept(Ret(std::move(*r)))
-> > ```
-> 
-> 3\. *Returns:* `std::move(*r)`
+> 1\. The name `iter_move` denotes a *customization point object* (17.5.2.1.5 [customization.point.object]). The effect of the expression `ranges::iter_move(E)` for some expression `E` is equivalent to `iter_move(E)`, with overload resolution performed in an unspecified context described below.
 >
-> <span style="color:blue">[*Editorial note:* -- Future work: Rather than defining a new `iter_swap` in namespace `std::experimental::ranges::v1`, it will probably be necessary to constrain the `iter_swap` in namespace `std` much the way the Ranges TS constrains `std::swap`. --*end note*]</span>
+> 2\. The context in which the expression `iter_move(E)` is evaluted
```
CaseyCarter: *evaluated*

ext/proxy/Dxxxx.md
```diff
@@ -1135,63 +1089,91 @@ Change 24.6 "Header `<experimental/ranges/iterator>` synopsis" ([iterator.synops
...
-> 1\. The return type is `Ret` where `Ret` is
-> `remove_reference_t<reference_t<I>>&&` if `I` is
-> a reference type; `decay_t<I>`, otherwise.
-> 2\. The expression in the `noexcept` is equivalent to:
-> > ```c++
-> > noexcept(Ret(std::move(*r)))
-> > ```
-> 
-> 3\. *Returns:* `std::move(*r)`
+> 1\. The name `iter_move` denotes a *customization point object* (17.5.2.1.5 [customization.point.object]). The effect of the expression `ranges::iter_move(E)` for some expression `E` is equivalent to `iter_move(E)`, with overload resolution performed in an unspecified context described below.
```
CaseyCarter: Is it unspecified, or is it described below?

ext/proxy/Dxxxx.md
```diff
@@ -1135,63 +1089,91 @@ Change 24.6 "Header `<experimental/ranges/iterator>` synopsis" ([iterator.synops
...
+> > > > `remove_reference_t<reference_t<I>>&&` if 
+> > > > `reference_t<I>` is a reference type;
+> > > > `decay_t<reference_t<I>>`, otherwise.
+> > > >
+> > > > 2.2. The expression in the `noexcept` is equivalent to:
+> > > > > ```c++
+> > > > > noexcept(Ret(std::move(*r)))
+> > > > > ```
+> > > >
+> > > > 2.3. *Effects:* Equivalent to `std::move(*r)`
```
CaseyCarter: `static_cast<Ret>(*r)` is preferable as per the resolution of [LWG 2106](wg21.link/lwg2106). When `*r` is a prvalue, `std::move(*r)` impedes move elision.

ext/proxy/Dxxxx.md
```diff
@@ -1135,63 +1089,91 @@ Change 24.6 "Header `<experimental/ranges/iterator>` synopsis" ([iterator.synops
...
 >
-> 4\. *Effects*: `swap(*r1, *r2)`
+> 3\. If the `iter_move` function selected by overload resolution ([over.match]) does not return a (possibly proxy) reference to, or copy of, the object denoted by `*r`, the program is ill-formed with no diagnostic required.
```
CaseyCarter: What is a "possibly proxy reference"? What does "copy of" mean? What if `*r` does not denote an object?

ext/proxy/Dxxxx.md
```diff
@@ -1135,63 +1089,91 @@ Change 24.6 "Header `<experimental/ranges/iterator>` synopsis" ([iterator.synops
...
-> >   requires !Swappable<reference_t<_R1>, reference_t<_R2>>()
-> >     && IndirectlyMovable<_R1, _R2>() && IndirectlyMovable<_R2, _R1>()
-> > void iter_swap(I1&& r1, I2&& r2)
-> >   noexcept(is_nothrow_indirectly_movable<_R1, _R2>::value &&
-> >            is_nothrow_indirectly_movable<_R2, _R1>::value);
-> > ```
+> 24.6.? *iter_swap*                     [iterator.utils.iter_swap]
 >
-> 5\. *Effects*: Exchanges values referred to by two `Readable` objects.
+> 1\. The name `iter_swap` denotes a *customization point object* (17.5.2.1.5 [customization.point.object]). The effect of the expression `ranges::iter_swap(E1, E2)` for some expressions `E1` and `E2` is equivalent to `(void)iter_swap(E1, E2)`, with overload resolution performed in an unspecified context described below:
```
CaseyCarter: Again, "unspecified" seems to contradict "described below."

ext/proxy/Dxxxx.md
```diff
@@ -1135,63 +1089,91 @@ Change 24.6 "Header `<experimental/ranges/iterator>` synopsis" ([iterator.synops
...
 >
-> 6\. \[*Example:* Below is a possible implementation:
-> > ```c++
-> > value_type_t<_R1> tmp(iter_move(r1));
-> > *r1 = iter_move(r2);
-> > *r2 = std::move(tmp);
-> > ```
+> 2\. The context in which the expression `(void)iter_swap(E1, E2)` is evaluted
```
CaseyCarter: s/evalute/evaluate/g =)

ext/proxy/Dxxxx.md
```diff
@@ -1135,63 +1089,91 @@ Change 24.6 "Header `<experimental/ranges/iterator>` synopsis" ([iterator.synops
...
+> > > > 2.3. \[*Example:* Below is a possible implementation:
+> > > > > ```c++
+> > > > > value_type_t<_R1> tmp(iter_move(r1));
+> > > > > *r1 = iter_move(r2);
+> > > > > *r2 = std::move(tmp);
+> > > > > ```
+> > > > >
+> > > > -- *end example*\]
+>
+> 3\. If the `iter_swap` function selected by overload resolution ([over.match]) does not swap the values of the objects to which the expressions `E1` and `E2` refer, the program is ill-formed with no diagnostic required.
```
CaseyCarter: proxy references don't necessarily refer to objects; I'd rephrase this. Maybe "swap the values denoted by the expressions `E1` and `E2`.

ext/proxy/Dxxxx.md
```diff
@@ -289,8 +289,8 @@ template <Readable I1, Readable I2>
   requires !Swappable<reference_t<I1>, reference_t<I2>>() &&
     IndirectlyMovable<I1, I2>() && IndirectlyMovable<I2, I1>()
```
CaseyCarter: These need to be `Storable` to support construction of `value_type_t<In>` from `rvalue_reference_t<In>`.

ext/proxy/Dxxxx.md
```diff
@@ -1135,63 +1089,91 @@ Change 24.6 "Header `<experimental/ranges/iterator>` synopsis" ([iterator.synops
...
+> > > ```
+> > > > 2.1. *Effects:* Equivalent to `ranges::swap(*r1, *r2)`
+> > >
+> > > ```c++
+> > > // Exposition only
+> > > template <class I1, class I2,
+> > >   Readable _R1 = std::remove_reference_t<I1>,
+> > >   Readable _R2 = std::remove_reference_t<I2>>
+> > >   requires !Swappable<reference_t<_R1>, reference_t<_R2>>()
+> > >     && IndirectlyMovable<_R1, _R2>() && IndirectlyMovable<_R2, _R1>()
```
CaseyCarter: Same here, these should be `Storable`.

ext/proxy/Dxxxx.md
CaseyCarter: Maybe just fall back to `iter_move(r)` must be equal to `*r`. Do I recall correctly that `iter_move(r)` is required to be equality-preserving?

ericniebler: I don't think we require that anywhere, but we should. How about:

> If the `iter_move` function selected by overload resolution ([over.match]) is not equality preserving, or if `iter_move(r)` does not equal `*r`, the program is ill-formed with no diagnostic required

ext/proxy/Dxxxx.md
ericniebler: good catch

ericniebler: > Am I understanding correctly that you're dropping the definitions of IndirectlyMovable at al. herein and referring to those in N4569?

Yes. The technical specifications reflects that but the preceding discussion does not. I can fix that.

ext/proxy/Dxxxx.md
CaseyCarter: We have `{ iter_move(i) } -> Same<rvalue_reference_t<I>>;` in `Readable`; that's enough to establish equality-preservation.

ericniebler: For the Readable concept yes. I thought you wanted the iter_move cust pt to offer that guarantee on its own.

ericniebler: I took that verbage out of the description of `iter_move`. I've incorporated all the rest of your feedback. This PR should be ready now.

PR #191, opened 2016-10-11
actually remove equality-preserving requirement on iter_move

ericniebler:
I thought I had committed these changes to the first PR but in fact I hadn't. Sorry.


CaseyCarter:
I decided my vague unease with using equality-preservation outside of a concept definition wasn't worth making an issue over. Feel free to commit or not as you like.

ericniebler:
We can always add it later. I prefer leaving it out. No other customization point says anything about equality preservation.

PR #192, opened 2016-10-11
s/\oldoldtxt/\removed/g

cjdb:
Notepad++ find/replace used.
If this looks okay (which I think it does), I'll do the same for s/\newnewtxt/\added/g.
\oldtxt and \newtxt to be done by hand only.


ericniebler:
Please put this on a branch. I don't want to commit to master half completed mark-up changes that don't represent a valid document state.

PR #195, opened 2016-10-11
Merge "Ranges TS Design Updates Omnibus"

CaseyCarter:
....to master.


ericniebler:
FWIW, I've avoided checking the acm csl file into github because I don't know the licensing of it.

CaseyCarter:
https://github.com/ericniebler/stl2/pull/195/files#diff-5f43abe7091ca61288a998753d548c4aR21 says it is CC-by-SA 3.0

PR #199, opened 2016-10-12
Markup reset on intro, lib-intro, numerics, and appendices

ericniebler:


lib-intro.tex
```diff
@@ -32,13 +32,13 @@
...
 
 \pnum
-Requirements are stated in terms of concepts~(Concepts TS [dcl.spec.concept]).
-Concepts are stated in terms of well-defined expressions that define valid terms of
-the types that satisfy the \oldoldtxt{requirements}\newnewtxt{concept}. For every set of
-well-defined expression requirements there is a \oldoldtxt{table}\newnewtxt{named concept} that
+Requirements are stated in terms of \added{concepts~(Concepts TS [dcl.spec.concept]).
+Concepts are stated in terms of} well-defined expressions that define valid terms of
+the types that satisfy the \changed{requirements}{concept}. For every set of
+well-defined expression requirements there is a \changed{table}{named concept} that
```
ericniebler: Notice that here we were changing standard text without any markup. Oops.

lib-intro.tex
```diff
@@ -32,13 +32,13 @@
...
 specifies an initial set of the valid expressions and their semantics. Any generic
 algorithm (Clause~\ref{algorithms}) that uses the well-defined expression requirements
-is described in terms of the valid expressions for its template type parameters.
+is described in terms of the valid expressions for its formal type parameters.
```
ericniebler: It's "template" in the latest draft, but "formal" in 4141. Reverting to the 4141 text.

future.tex
```diff
@@ -150,10 +150,8 @@
 instance. There may be other useful clusters of concepts for other algorithm groups. Finding them
 and using them to improve the algorithm constraints is an important work item.
 
-{\color{newclr}
 \rSec1[future.initlsts]{Initializer Lists}
```
CaseyCarter: We should just cut this and hope no one notices.

PR #213, opened 2016-10-13
Markup reset

ericniebler:


CaseyCarter:
Wow that's a big diff. I don't think we need to review it again.

ericniebler:
Bombs away. @CaseyCarter, I cherry-picked your non-markup-related changes so they wouldn't get lost in the history with the squash-and-merge. Thanks, all!

CaseyCarter:
> I cherry-picked your non-markup-related changes so they wouldn't get lost in the history with the squash-and-merge.

Yes, thanks. There will be a reckoning when we finally "catch up" the true Working Paper to our process document and need to produce a full detailed editor's report.

PR #217, opened 2016-10-14
Cleanup and integration of P0441

### LINKED ISSUE #197: Why does `IndirectlyCopyable` refine `IndirectlyMovable`
...instead of being defined as:

```c++
template <class In, class Out>
concept bool IndirectlyCopyable() {
  return Readable<In>() && Writable<Out, reference_t<In>>();
}
```

certainly there are no algorithms that *both* move and copy elements from their input range to their output range. I write either `*out = *in;` or `*out = iter_move(*in);`, never both.

CaseyCarter:
Closes #197.

### DIFF
--- a/ext/writable/P0441.md
+++ b/ext/writable/P0441.md
@@ -5,7 +5,7 @@ title: 'Ranges: Merging Writable and MoveWritable'
 
 # Introduction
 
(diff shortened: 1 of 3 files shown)


ext/writable/P0441.md
```diff
@@ -455,1079 +463,13 @@ In [alg.partitions], strike the declarations of `partition_move` and remove para
...
-Replace the contents of [iterators.common], [default.sentinels], [iterators.counted], and [unreachable.sentinels] with the contents of the sections of the same name in N4569.
-
-Strip occurrences of the prefix `Weak` wherever it appears in clause [algorithms].
-
-# Acknowledgements
-The authors would like to thank Andrew Sutton and Sean Parent for their participation in the discussions that produced most of the ideas herein.
--->
-
-### Implementation Experience
-The proposed design changes are implemented in both [CMCSTL2, a full implementation of the Ranges TS with proxy extensions][2][@cmcstl2], and [range-v3][5][@range-v3].
```
ericniebler: Could you explain why enormous blocks of the doc have been deleted?

ericniebler:
We should probably stop committing the pdf file to the repo.

CaseyCarter:
> We should probably stop committing the pdf file to the repo.

At least for now.

PR #218, opened 2016-10-14
integrate Jacksonville review feedback

ericniebler:
Pretty simple changes.

### DIFF
--- a/iterators.tex
+++ b/iterators.tex
@@ -202,7 +202,7 @@ and up to but not including the element pointed to by
 \added{the first iterator} \tcode{j} \added{such that \tcode{j == s}}.
 
 \pnum
-Range \removed{\range{i}{j}}\added{\range{i}{s}}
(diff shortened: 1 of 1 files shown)


iterators.tex
```diff
@@ -339,8 +344,8 @@
   template <class Out, class T>
   concept bool Writable() {
     return Semiregular<Out>() &&
-      requires(Out o, T&& v) {
-        *o = std::forward<T>(v); // not required to be equality preserving
+      requires(Out o, T&& t) {
+        *o = std::forward<T>(t); // not required to be equality preserving
```
CaseyCarter: These changes in `Writable` are causing the conflict: I already replaced `v` with `t` in the concept definition, and I replaced `v` in the text with `E`, "an expression such that `decltype(E)` is `T`".

ericniebler:
Assuming you had no other comments, I'm committing.

PR #222, opened 2016-10-14
describes->specifies, more changes from ranges telecon 3

ericniebler:
- The standard doesn't describe. It *specifies*.
- Replace "comparable for equality" with "satisfy EqualityComparable" in description (specification) of InputIterator.

### DIFF
--- a/iterators.tex
+++ b/iterators.tex
@@ -336,7 +336,7 @@ like \tcode{shared_ptr<void>} is not \tcode{Readable} and has no associated valu
 \rSec2[iterators.writable]{Concept Writable}
 
 \pnum
-The \tcode{Writable} concept describes the requirements for writing a value into an iterator's
+The \tcode{Writable} concept specifies the requirements for writing a value into an iterator's
 referenced object.
 
 \indexlibrary{\idxcode{Writable}}%
@@ -382,8 +382,9 @@ The only valid use of an \tcode{operator*} is on the left side of the assignment
 \rSec2[iterators.weaklyincrementable]{Concept WeaklyIncrementable}
 
 \pnum
-The \tcode{WeaklyIncrementable} concept describes types that can be incremented with the pre-
-and post-increment operators. The increment operations are not required to be equality-preserving,
+The \tcode{WeaklyIncrementable} concept specifies the requirements on
+types that can be incremented with the pre- and post-increment operators.
+The increment operations are not required to be equality-preserving,
 nor is the type required to be \tcode{EqualityComparable}.
 
 \indexlibrary{\idxcode{WeaklyIncrementable}}%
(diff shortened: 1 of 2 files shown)


iterators.tex
```diff
@@ -408,7 +409,7 @@
 \begin{itemize}
 \item \tcode{++i} is valid if and only if \tcode{i++} is valid.
 \item If \tcode{i} is incrementable, then both \tcode{++i}
-  and \tcode{i++} advance \tcode{i} to the next element.
+  and \tcode{i++} advance \tcode{i} to denote the next (possibly past-the-end) element.
```
CaseyCarter: I'll reiterate my position that I despise this meaningless requirement and think it should be struck. What's an "element"? What does "denote" mean here? What's a "next element"? What is a "past-the-end" element? These terms have meaning in the context of iterators, but this concept - despite being (mis?) located in iterator.requirements - is not an iterator concept.

At the very least, we should relocate this requirement to `Iterator`.

iterators.tex
```diff
@@ -608,7 +609,7 @@
 \tcode{Iterator}~(\ref{iterators.iterator}). It
 defines requirements for a type whose referenced values can be read (from the requirement for
 \tcode{Readable}~(\ref{iterators.readable})) and which can be both pre- and post-incremented. However,
-input iterators are not required to be comparable for equality.
+input iterators are not required to satisfy \tcode{EqualityComparable}~(\ref{concepts.lib.compare.equalitycomparable}).
```
CaseyCarter: "However" doesn't make sense to me here. How about:
```tex
\enternote Unlike in ISO/IEC 14882, input iterators are not required to
satisfy \tcode{EqualityComparable}. \exitnote
```

ericniebler: That raises the issue in my mind of what `EqualityComparable` means in ISO/IEC 14882, and how it is related to what `EqualityComparable` means in this document, but maybe that's OK. Also, I think this should be a note, since the lack of the `EqualityComparable` requirement is already specified in the `InputIterator` concept definition.

iterators.tex
ericniebler: Oh wow, you're right. Gimme a sec.

CaseyCarter: They aren't equality-preserving, so we can't even require pre and post increment to have the same (unspecified) effect.

CaseyCarter: Do we even *need* `WeaklyIncrementable`, or can we replace all uses with `Iterator`?

CaseyCarter: > Do we even need `WeaklyIncrementable`, or can we replace all uses with `Iterator`?

Paraphrase: Are there models of `WeaklyIncrementable` that do not model `Iterator`? Certainly I can imagine models of `Incrementable` that do not model `Iterator` (`int` comes to mind), but non-`Iterator` things with-non-EP-increment seem rare in the standard library.

CaseyCarter: > I'm inclined to just back this change out for now and come back to the committee with a proposal when we'v thought more deeply about it.

Agreed. I know I've had this "remove `WeaklyIncrementable`" argument with myself before, and there doesn't seem to be a good choice.

ericniebler: `view::iota` uses `WeaklyIncrementable`. I think it's generally useful.

CaseyCarter: I suspect it's only useful as the intersection of `Incrementable` and `Iterator`, i.e., a cheaper-to-compile `Incrementable<T>() || Iterator<T>()`. We'll see ;)

ericniebler:
Reverted the WeaklyIncrementable op++ change. Will revisit later.

PR #224, opened 2016-10-17
Memory proposal

cjdb:
Added `uninitialized_copy`, `uninitialized_copy_n`, `uninitialized_fill`, and `uninitialized_fill_n` to the draft. Should I remove the oldtxt/newtxt markup, and can I add C++17 algorithms, such as uninitialized_move, destroy, etc., or leave it as-is?


cjdb:
Added `destroy` and `destroy_n` from the C++17 working draft in a separate branch @c9bfb2d9b5780c487ecc5f5d725f7538116fee1c as an example.

ericniebler:
We shouldn't be making new additions directly to the Ranges TS. That we have done it in the past is due to Casey's and my confusion about the TS process. New changes have to go in a separate paper and only integrated when we get full committee sign-off.

I'm also wary of expanding the scope of the Ranges TS at this point. I don't want anything to delay it. We need to think of how best to use the committee's limited time; time spent reviewing the memory algorithms is time taken away from finishing the TS. At some point, it will be time to start thinking of Ranges TS2, and making that the delivery vehicle for additional functionality. See Library Fundamentals TS1 & 2, for example.

ericniebler:
tl;dr please start a new proposal for the memory algorithms. It doesn't have to be elaborate.

PR #228, opened 2016-10-18
[Editorial] Strike "purely semantic" note about (Weakly)?EqualityComparable

### LINKED ISSUE #201: EqualityComparable vs WeaklyEqualityComparable
The text has this:

> "[ Note: The distinction between EqualityComparable&lt;T, U&gt;() and WeaklyEqualityComparable&lt;T, U&gt;() is purely semantic. -- end note ]"

Really? `EqualityComparable` requires `Common<T,U>` and `WeaklyEqualityComparable` doesn't. That's more that semantic, isn't it?

CaseyCarter:
Fixes #201.

### DIFF
--- a/concepts.tex
+++ b/concepts.tex
@@ -546,12 +546,6 @@ is satisfied if and only if:
 \end{itemize}
 \end{itemdescr}
 
-\pnum
-\enternote The distinction between
-\tcode{Equal\-ity\-Comp\-arable<\brk{}T, \brk{}U>()} and
-\tcode{Weakly\-Equal\-ity\-Comp\-arable<T, U>()} is purely semantic.
-\exitnote
-
 \rSec2[concepts.lib.compare.stricttotallyordered]{Concept StrictTotallyOrdered}
 
 \ednote{Remove table [lessthancomparable] in [utility.arg.requirements]. Replace uses of


ericniebler:
Hang on. Not sure I'd call this editorial. Besides, we need to stop editing the TS directly, and start a new doc with all the fixes we want. We can submit that doc for review at Kona in February.

CaseyCarter:
Notes are non-normative. Does removing a note not fall within the purview of a purely editorial change?

ericniebler:
I don't actually know. @jwakely, can you weigh in on the process here?

1. Is it OK to make editorial changes to the working draft of a TS
  - without the changes appearing first in a separate paper that is reviewed by LWG? and
  - without change markings?
2. Is removing an erroneous note editorial?

jwakely:
1. Yes, it's definitely OK to make editorial changes without them going through the committee (I've made dozens to the Networking TS, to improve formatting, make variable names more consistent, add index entries and so on). I'm also not keeping change marks in the Networking TS, just noting any significant changes in the Editor's Report document that goes in the mailing alongside a new draft.

2. I believe so. Notes are not normative, they just help to clarify the normative text, so if it's failing to clarify anything (because it's wrong!) then I think removing it is editorial.

There's quite a good definition of what qualifies as editorial at https://github.com/cplusplus/draft/wiki/How-to-tell-if-an-issue-is-editorial and that says any change to non-normative text can be done by the editor.

CaseyCarter:
Thanks for the clarification, Jonathan.

ericniebler:
Thanks, @jwakely!

PR #265, opened 2016-11-14
Integration of P0370 and P0441

CaseyCarter:
...with editorial changes from Issaquah LWG review.


CaseyCarter:
There's a fresh pull without the redundant markup change. I have no idea where all the comments went. (Perhaps they were attached to the old PR, and I simply didn't notice.)

PR #272, opened 2016-11-15
add missing front matter to utilities clause, misc fixups and review …

ericniebler:
…feedback

### DIFF
--- a/utilities.tex
+++ b/utilities.tex
@@ -8,7 +8,6 @@ This Clause describes utilities that are generally useful in \Cpp programs; some
 of these utilities are used by other elements of the Ranges library.
 These utilities are summarized in Table~\ref{tab:util.lib.summary}.
 
-\setcounter{table}{2}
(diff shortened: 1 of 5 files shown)


utilities.tex
```diff
@@ -1,12 +1,36 @@
...
 \rSec0[utilities]{General utilities library}
 
+\rSec1[utilities.general]{General}
+
+\pnum
+This Clause describes utilities that are generally useful in \Cpp programs; some
+of these utilities are used by other elements of the Ranges library.
+These utilities are summarized in Table~\ref{tab:util.lib.summary}.
+
+\setcounter{table}{2}
```
CaseyCarter: Why are we messing with the table counter here?

CaseyCarter: There are no tables in Clause 2 (Statements). There are correctly numbered tables in Clauses 3 (Table 2) and 4 (Table 3), followed by this incorrectly numbered Table 3.

ericniebler: Sorry, fixed.

PR #273, opened 2016-11-15
Remove space in definition of cxxref and stdcxxref

CaseyCarter:


ericniebler:
Thanks. I couldn't figure that out.

PR #275, opened 2016-11-15
add missing sections and fix bogus references to meta stuff

ericniebler:


CaseyCarter:
I'm not enthused about the ednotes, but this will do until we come up with something better.

Issue #210, opened 2016-10-13
Insert iterators still have "typename Container::iterator" in their interface


ericniebler:
g/`typename Container::iterator`/`iterator_t<Container>`/g

Ditto for `typename Container::value_type`.

### Proposed Resolution:

Change the `insert_iterator` class definition ([insert.iterator]) as follows:

```diff
template <class Container>
class insert_iterator {
private:
  Container* container;              // \expos
-  typename Container::iterator iter; // \expos
+  iterator_t<Container> iter; // \expos

public:
  using container_type = Container;
  using difference_type = ptrdiff_t;
  insert_iterator();
-  insert_iterator(Container& x, typename Container::iterator i);
+  insert_iterator(Container& x, iterator_t<Container> i);
  insert_iterator&
-   operator=(const typename Container::value_type& value);
+   operator=(const value_type_t<Container>& value);
  insert_iterator&
-   operator=(typename Container::value_type&& value);
+   operator=(value_type_t<Container>&& value);
```

Make the corresponding changes in the member function definitions.

Change the `back_insert_iterator` class definition ([back.insert.iterator]) as follows:

```diff
explicit back_insert_iterator(Container& x);
back_insert_iterator&
-  operator=(const typename Container::value_type& value);
+  operator=(const value_type_t<Container>& value);
back_insert_iterator&
-  operator=(typename Container::value_type&& value);
+  operator=(value_type_t<Container>&& value);
```

Make the corresponding changes to the member function definitions.

Change the `front_insert_iterator` class definition ([front.insert.iterator]) as follows:

```diff
explicit front_insert_iterator(Container& x);
front_insert_iterator&
-  operator=(const typename Container::value_type& value);
+  operator=(const value_type_t<Container>& value);
front_insert_iterator&
-  operator=(typename Container::value_type&& value);
+  operator=(value_type_t<Container>&& value);
```

Make the corresponding changes to the member function definitions.

CaseyCarter:
Closed by ebf0fef9b76d900d11a4fc0512e7ad0ca5335463.

PR #279, opened 2016-11-16
This markup shall not stand: part 4

CaseyCarter:


iterators.tex
```diff
@@ -2991,41 +2986,39 @@
...
-    @\changed{typedef Container}{using}@ container_type@\added{ = Container}@;
-    @\added{using difference_type = ptrdiff_t;}@
-    @\added{insert_iterator();}@
-    insert_iterator(Container& x, typename Container::iterator i);
-    insert_iterator@\removed{<Container>}@&
+    using container_type = Container;
+    using difference_type = ptrdiff_t;
+
+    insert_iterator();
+    insert_iterator(Container& x, iterator_t<Container> i);
```
CaseyCarter: I'm editorially interpreting that the intent of the change from `typename Container::iterator` to `iterator_t<Container>` was to make that change throughout `insert_iterator`, and not only to the few places that were actually changed,

ericniebler: Actually, it was to brink the text body in line with a change that had already been made in the synopsis. I held off on the the other edits until LWG approved the change.

iterators.tex
```diff
@@ -3777,7 +3695,7 @@
 is a \tcode{constexpr} constructor.
 \end{itemdescr}
 
-\indexlibrary{\idxcode{move_iterator}!constructor}%
+\indexlibrary{\idxcode{move_sentinel}!constructor}%
```
ericniebler: Nice catch.

ericniebler:
Fixes #210

iterators.tex
```diff
@@ -3064,13 +3053,13 @@
 
 \indexlibrary{\idxcode{operator=}!\idxcode{insert_iterator}}%
 \begin{itemdecl}
-insert_iterator@\removed{<Container>}@&
+insert_iterator&
   operator=(const typename Container::value_type& value);
```
CaseyCarter: I think I can call this editorial with a straight face.

Issue #211, opened 2016-10-13
Add new header <experimental/range/range>


ericniebler:
Where do `enable_view`, `disable_sized_range`, and `view_base` live? Or the Range concepts? Do we need a `<experimental/ranges/range>` header?

Heck, where do the Iterator concepts live? They don't appear in the `iterator` header synopsis.

Solution description
--------------------

The problems with the concepts and iterator synopses have been handled as editorial changes in 1862bd8 and e15540b. This resolution is describes as a diff to those changes. It suggests adding a new header to hold all range-related functionality: `<experimental/ranges/range>`. This gives us a place to put the range customization points (`begin`, `end`, `size`, etc.), the range concepts, and range utilities.

It will also give us a place to put future range-related functionality, such as a view facade and adaptors, without further bloating `<experimental/ranges/iterator>`, which is quite heavy already.

The following resolution has been implemented in #328, but not merged pending LWG review.

Proposed Resolution
--------------------

To "Table 1 - Ranges TS library headers", add `<experimental/ranges/range>`.

To "Table 2 - Library categories", add a row for "Ranges library" between the "Iterators library" and the "Algorithms library".

After [library.general]/p6, add a new paragraph that reads:

> The ranges library (\ref{ranges}) describes components for dealing with ranges of
> elements.

Change [iterators.general]/1 as follows:

```diff
This Clause describes components that \Cpp programs may use to perform
iterations over containers (Clause \cxxref{containers}),
streams~(\cxxref{iostream.format}),
-stream buffers~(\cxxref{stream.buffers}),
-and ranges~(\ref{ranges}).
+and stream buffers~(\cxxref{stream.buffers}).
```

From table "Table 5 — Iterators library summary" [tab:iterators.lib.summary], delete the last line, which reads "Ranges".

From [iterator.synopsis], delete the following lines:

```diff
-  // \ref{iterator.range}, range access:
-  namespace {
-    constexpr @\unspec@ begin = @\unspec@;
-    constexpr @\unspec@ end = @\unspec@;
-    constexpr @\unspec@ cbegin = @\unspec@;
-    constexpr @\unspec@ cend = @\unspec@;
-    constexpr @\unspec@ rbegin = @\unspec@;
-    constexpr @\unspec@ rend = @\unspec@;
-    constexpr @\unspec@ crbegin = @\unspec@;
-    constexpr @\unspec@ crend = @\unspec@;
-  }
-
-  // \ref{range.primitives}, range primitives:
-  namespace {
[...]

PR #324, opened 2017-02-14
[EDITORIAL] add <experimental/ranges/concepts> synopsis

ericniebler:


ericniebler:
Refs #211

concepts.tex
```diff
@@ -127,6 +127,209 @@
 which requires \tcode{C<T>()}.
 \exitexample
 
+\rSec1[concepts.lib.synopsis]{Header \tcode{<experimental/ranges/concepts>} synopsis}
```
CaseyCarter: With the synopsis, [concepts.lib.general]/3 is now redundant and should be removed.

concepts.tex
```diff
@@ -127,6 +127,209 @@
...
+
+\indexlibrary{\idxhdr{experimental/ranges/concepts}}%
+\begin{codeblock}
+namespace std { namespace experimental { namespace ranges { inline namespace v1 {
+  // \ref{concepts.lib.corelang}, core language concepts:
+  // \ref{concepts.lib.corelang.same}, Same:
+  template <class T, class U>
+  concept bool Same() {
```
CaseyCarter: Should we add another indent on these to match the standard's function template declaration style?

ericniebler: We don't elsewhere for concept definitions. That change is not in the scope of this patch.

Issue #284, opened 2016-11-17
iter_move and iter_swap need to say when they are noexcept and constexpr


ericniebler:
The proposed wording of #242 takes a stab at it but gets it wrong.

Proposed Resolution
-----------------------

Add two new paragraphs to [intro.defs] that read:

> -?-
> **constant subexpression**
> expression whose evaluation as subexpression of a conditional-expression `CE` (ISO/IEC 14882:2014 §5.16) would not prevent `CE` from being a core constant expression (ISO/IEC 14882:2014 §5.19)
>
> -?-
> **expression-equivalent**
> relationship that exists between two expressions `E1` and `E2` such that
> - `E1` and `E2` have the same effects,
> - `noexcept(E1) == noexcept(E2)`, and
> - `E1` is a constant subexpression if and only if `E2` is a constant subexpression

[_Editor's note:_ the definition of "constant subexpression" is taken from the latest C++17 draft.]

Change [iterator.utils.iter_move]/1 as follows:

```diff
 1 The name iter_move denotes a customization point
-  object (3.3.2.3). The effect of the expression
+  object (3.3.2.3). The expression
   ranges::iter_move(E) for some expression E is
-  equivalent to the following:
+  expression-equivalent to the following:
-(1.1) — iter_move(E), if that expression is well-formed
+(1.1) — static_cast<decltype(iter_move(E))>(iter_move(E)),
+      if that expression is well-formed
       when evaluated in a context that does not include
       ranges::iter_move but does include the lookup set
       produced by argument-dependent lookup (ISO/IEC
       14882:2014 §3.4.2).
-(1.2) — Otherwise, if the expression *E is well-formed
-(1.2.1) — If *E is an lvalue, std::move(*E).
-(1.2.2) — Otherwise, *E.
+(1.2) — Otherwise, if the expression *E is well-formed:
+(1.2.1) — if *E is an lvalue, std::move(*E);
+(1.2.2) — otherwise, static_cast<decltype(*E)>(*E).
 (1.3) — Otherwise, ranges::iter_move(E) is ill-formed.
```

Change [iterator.utils.iter_swap]/1 as follows:

```diff
 1 The name iter_swap denotes a customization point
-  object (3.3.2.3). The effect of the expression
+  object (3.3.2.3). The expression
   ranges::iter_swap(E1, E2) for some expressions E1
-  and E2 is equivalent to the following:
+  and E2 is expression-equivalent to the following:
[...]

PR #326, opened 2017-02-14
[EDITORIAL] major reorg of <iterator>; add concepts to the synopsis

ericniebler:
This complicates the application of #284. Where does `iter_exchange` go? It can't go with `iter_move` and `iter_swap` since it isn't a customization point. And if `iter_swap` is specified in terms of `iter_exchange`, then it is making use of concepts that haven't been defined yet. Even so, I feel this is a (vast) improvement.

### DIFF
--- a/iterators.tex
+++ b/iterators.tex
@@ -29,1063 +29,242 @@ as summarized in Table~\ref{tab:iterators.lib.summary}.
 \ref{ranges}                & Ranges                         &                           \\
 \end{libsumtab}
 
-\rSec1[iterator.requirements]{Iterator requirements}
-
-\rSec2[iterator.requirements.general]{In general}
-
-\pnum
-\indextext{requirements!iterator}%
-Iterators are a generalization of pointers that allow a \Cpp program to work with different data structures
-(for example, containers and ranges) in a uniform manner.
-To be able to construct template algorithms that work correctly and
-efficiently on different types of data structures, the library formalizes not just the interfaces but also the
-semantics and complexity assumptions of iterators.
-All input iterators
-\tcode{i}
-support the expression
-\tcode{*i},
-resulting in a value of some object type
-\tcode{T},
-called the
-\term{value type}
-of the iterator.
-All output iterators support the expression
-\tcode{*i = o}
-where
-\tcode{o}
-is a value of some type that is in the set of types that are
-\term{writable}
-to the particular iterator type of
-\tcode{i}.
-For every iterator type
-\tcode{X}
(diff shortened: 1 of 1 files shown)


ericniebler:
Actually, I don't think it's a problem. `iter_swap` only uses `iter_exchange` in a "expression-equivalent" way; there's no requirement that `iter_swap` _actually_ use `iter_exchange`. We can put `iter_exchange` farther down in the synopsis, perhaps in [iterator.operations] after `next` and `prev`. It would be a bit odd to not have `iter_exchange` with `iter_move` and `iter_swap`, but it really is a different beast. It is not a customization point object.

iterators.tex
```diff
@@ -29,1674 +29,1871 @@
...
-\tcode{difference_type_t<I>} such that after
-\tcode{n} applications of \tcode{++a}, then \tcode{bool(a == b)}. Then
-\tcode{RandomAccessIterator<I>()} is satisfied if and only if:
+Since iterators are an abstraction of pointers, their semantics are
+a generalization of most of the semantics of pointers in \Cpp.
+This ensures that every
+function template
+that takes iterators
+works as well with regular pointers.
+This International Standard defines
```
CaseyCarter: s/International Standard/document/

iterators.tex
```diff
@@ -29,1674 +29,1871 @@
...
-For every iterator type
-\tcode{X}
-there is a corresponding signed integer type called the
-\term{difference type}
-of the iterator.
+\indexlibrary{\idxhdr{experimental/ranges/iterator}}%
+\begin{codeblock}
+namespace std { namespace experimental { namespace ranges { inline namespace v1 {
+  template <class T> concept bool @\placeholder{dereferenceable}@ // \expos
+    = requires(T& t) { {*t} -> auto&&; };
```
CaseyCarter: Needs more spaces in the braces: `{ *t } -> auto&&;`.

iterators.tex
```diff
@@ -29,1674 +29,1871 @@
...
-    requires is_array<I>::value
-  struct value_type<I> : value_type<decay_t<I>> { };
+  concept bool BidirectionalIterator() {
+    return @\seebelow@;
+  }
 
+  // \ref{iterators.random.access}, RandomAccessIterator:
   template <class I>
-  struct value_type<I const> : value_type<decay_t<I>> { };
+  concept bool BidirectionalIterator() {
```
CaseyCarter: Nope.

CaseyCarter:
This is a huge improvement; well done.

PR #328, opened 2017-02-15
refactor range-related things into new <experimental/ranges/range> header

ericniebler:
*********DO NOT MERGE*********

This is a test to see if these changes are an improvement or not. I think they are. @CaseyCarter what do you think?


ranges-lib.tex
```diff
@@ -0,0 +1,797 @@
...
+  difference_type_t<iterator_t<R>> distance(R&& r);
+}}}}
+\end{codeblock}
+
+\rSec1[range.access]{Range access}
+
+\pnum
+In addition to being available via inclusion of the \tcode{<experimental/ranges/range>}
+header, the customization point objects in \ref{range.access} are
+available when \tcode{<experimental/ranges/iterator>} is included
```
CaseyCarter: This sentence is missing a period at the end

ranges-lib.tex
```diff
@@ -0,0 +1,797 @@
...
+\enternote Whenever \tcode{ranges::crend(E)} is a valid expression, the
+types of \tcode{ranges::crend(E)} and \tcode{ranges::crbegin(E)} satisfy
+\tcode{Sentinel}. \exitnote
+
+\rSec1[range.primitives]{Range primitives}
+
+\pnum
+In addition to being available via inclusion of the \tcode{<experimental/ranges/range>}
+header, the customization point objects in \ref{range.primitives} are
+available when \tcode{<experimental/ranges/iterator>} is included
```
CaseyCarter: Missing period from this cut'n'paste copy of the prior sentence as well

ericniebler:
Given the before and after, I _vastly_ prefer having the separate `<experimental/ranges/range>` header.

ranges-lib.tex
```diff
@@ -0,0 +1,797 @@
...
+  using iterator_t = decltype(ranges::begin(declval<T&>()));
+
+  template <class T>
+  using sentinel_t = decltype(ranges::end(declval<T&>()));
+
+  template <class>
+  constexpr bool disable_sized_range = false;
+
+  template <class T>
+  struct enable_view { };
```
CaseyCarter: WTF `enable_view` is still a class template? Any reason we shouldn't make this a `constexpr bool` variable template like the other user-customizable predicates?

ericniebler: Because enable_view is tri state. Empty struct has meaning.

CaseyCarter: Second question: Why do I ask this same question about `enable_view` every 6-8 months?

ericniebler:
Refs #211

CaseyCarter:
FYI: rebased onto master after merging rebased-issaquah-issues-processing.

ericniebler:
Mini-LWG wanted to revisit the issue of where the range customization points (`begin`, `end`, etc.) live. It raised red flags that they were (a) specified to live in `<...range>`, (b) also available via `<...iterator>`, and (c) that `<...range>` was specified to `#include` <...iterator>`.

ericniebler:
Superseded by #442 .

PR #373, opened 2017-03-09
Define "projection"; Fixes #221.

### LINKED ISSUE #221: Define "projection"
...and reference it from [projected].

CaseyCarter:
Despite being editorial, this is probably interesting enough to merit more eyes than just mine.

### DIFF
--- a/intro.tex
+++ b/intro.tex
@@ -83,6 +83,25 @@ ISO/IEC 14882:2014 is herein called the \defn{\Cpp Standard}, N3351 is called th
 Terms defined in \cxxref{definitions} are used in this document with the same
 meaning.
 
+\def\definition{\definitionx{\section}}%
+
+\indexdefn{projection}%
+\definition{projection}{defns.projection}
+\defncontext{function object argument} used by an algorithm to examine the
+values of elements. This enables algorithms to behave based on the projected
+values rather than the actual element values.
+
+\enterexample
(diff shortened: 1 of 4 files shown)


intro.tex
```diff
@@ -83,6 +83,25 @@
 Terms defined in \cxxref{definitions} are used in this document with the same
 meaning.
 
+\def\definition{\definitionx{\section}}%
+
+\indexdefn{projection}%
+\definition{projection}{defns.projection}
+\defncontext{function object argument} which an algorithm uses to examine the
```
ericniebler: s/which/that/

CaseyCarter:
"that" vs. "which" is the bane of my existence. Better?

ericniebler:
> \defncontext{function object argument} used by an algorithm to examine the values of elements indirectly.

I don't love the use of the word "indirectly". So far, we've been using "indirectly" to mean "through a `Readable`". The projection doesn't get passed a `Readable`, it gets passed a value. I suggest striking "indirectly".

ericniebler:
> "that" vs. "which" is the bane of my existence.

Try mentally excising the clause that begins with "which". If you can't do it without changing the meaning of the sentence, replace "which" with "that".

CaseyCarter:
Killed "indirectly" in the definition of "projection".

Issue #299, opened 2016-12-05
`value_type` of classes with member `element_type`


CaseyCarter:
`element_type` in `std::pointer_traits` and in the smart pointers is the type of the object the pointer denotes, not the value type of that object (i.e., it can be a cv-qualified type). The specialization of `value_type` that uses `typename T::element_type`  should strip cv-qualifiers from the type.

### Proposed Resolution:
Modify [iterator.assoc.types.value_type] as follows:

```diff
 [...]
 
 2 A Readable type has an associated value type that can be accessed with the value_type_t alias
   template.

     [...]
     template <class T>
       requires requires { typename T::value_type; }
     struct value_type<T>
       : enable_if<is_object<typename T::value_type>::value, typename T::value_type> { };
 
     template <class T>
       requires requires { typename T::element_type; }
     struct value_type<T>
-      : enable_if<is_object<typename T::element_type>::value, typename T::element_type> { };
+      : enable_if<
+          is_object<typename T::element_type>::value,
+          remove_cv_t<typename T::element_type>>
+    { };
 
     template <class T>
       using value_type_t = typename value_type<T>::type;

 [...]

 5 When instantiated with a type I such that I::value_type is valid and denotes a type, [...]
 
 6 When instantiated with a type I such that I::element_type is valid and denotes a type,
-  value_type<I>::type names that type,
+  value_type<I>::type names the type remove_cv_t<I::element_type>,
   unless it is not an object type (ISO/IEC 14882:2014 §3.9)
   in which case value_type<I> shall have no nested type type. [ Note: Smart pointers like
   shared_ptr<int> are Readable and have an associated value type. But a smart pointer like
   shared_ptr<void> is not Readable and has no associated value type.—end note ]
```

CaseyCarter:
LWG: Add a note explaining the `remove_cv_t` for `element_type`.

cjdb:
Did this section move? It's under [iterator.assoc.types.value_type] for me.

cjdb:
If not, should it be moved to [iterators.readable]?

viboes:
Sorry for the question, but could someone explain why do we need to mix `requires` clauses with `enable_if`? 
Why the `is_object<typename T::value_type>::value` is not part of the requires?

ericniebler:
> Sorry for the question, but could someone explain why do we need to mix requires clauses with enable_if?

I can imagine a case where it matters:

```c++
struct S {
  using value_type = void;
  using element_type = int;
};
```

[...]

PR #378, opened 2017-03-10
Patch for #299 (`value_type` of classes with member `element_type`)

cjdb:


ericniebler:
Looks fine. We won't be able to merge this to master until the resolution is moved in plenary in Toronto.

PR #385, opened 2017-03-15
[EDITORIAL] LWG Kona changes:

CaseyCarter:
* Use decltype(auto) instead of "see below" in declarations

* Insert missing semicolons in the specification of counted_iterator::operator++

* Simplify counted_iterator::operator++(int) by using pre-increment in the effects-equivalent-to, which notably "inherits" the Requires element.


CaseyCarter:
This needs a revision history to indicate the differences with R0.

We also still need to address my note:

> basic guarantee in `counted_iterator`'s postincrement and postdecrement. (If that results in inconsistently ordered operations, STL wants a note to explain why.)

PR #395, opened 2017-03-26
Use ConvertibleTo to define Common and CommonReference (#311)

### LINKED ISSUE #311: Common and CommonReference should use ConvertibleTo to test for implicit convertibility
**Issue # 1:** the expressions `common_type_t<T, U>(t())` and `common_reference_t<T, U>(t())` in the `Common` and `CommonReference` concepts could potentially be interpreted as function-style casts depending on the types involved, which could amount to `reinterpret_cast`s. That was not the intention.

The intention was to test that expressions of type `T` and `U` can be converted to the common type (or the common reference). The fix is to simply test that requirement instead with `ConvertibleTo`.

**Issue # 2:** One of the potential uses for `common_`(`type`|`reference`) is to have a way to declare monomorphic functions that satisfy e.g., `IndirectRelation` for use with the std:: algorithms. Indeed, `IndirectReference` requires the invocable to be callable with common-reference parameters. Consider:

```c++
auto rng = view::zip(vec1, vec2); // from range-v3
using R = iter_common_reference_t<iterator_t<decltype(rng)>>;
// Note: R might be different than decltype(*begin(rng)) for, e.g. a proxy iterator.
auto pred = [](R a, R b) { return a < b; };
ranges::sort( rng, pred );
```

Right now all the higher order functions are more or less broken because they just do `pred(*here, *there)`. For an iterator whose reference type is explicitly-but-not-implicitly convertible to the common reference type, the call will pass the concept check but fail to instantiate. That's Bad.

As above, the fix is to change `Common` and `CommonReference` to use `ConvertibleTo`. That requires _implicit_ convertibility, rather than the explicit convertibility that was being checked with the function-style cast.

_Note:_ The underlying traits `common_type` and `common_reference` only require explicit convertibility. We think it's fine, indeed essential for generic code, for the concepts to be more strict than the underlying type traits.

### Proposed Resolution

Change the definition of `CommonReference` in [concepts.lib.corelang.commonref]/1 as follows:

```diff
 template <class T, class U>
 concept bool CommonReference() {
-  return requires(T (&t)(), U (&u)()) {
-    typename common_reference_t<T, U>;
-    typename common_reference_t<U, T>;
-    requires Same<common_reference_t<T, U>, common_reference_t<U, T>>();
-    common_reference_t<T, U>(t());
-    common_reference_t<T, U>(u());
-  };
+  return
+    Same<common_reference_t<T, U>, common_reference_t<U, T>>() &&
+    ConvertibleTo<T, common_reference_t<T, U>>() &&
+    ConvertibleTo<U, common_reference_t<T, U>>();
 }
```

*Note:* this removes the `typename common_reference_t...` constraints since they are superfluous.

Change the definition of `Common` in [concepts.lib.corelang.common]/1 as follows:

```diff
 template <class T, class U>
 concept bool Common() {
-   return CommonReference<const T&, const U&>() &&
-    requires(T (&t)(), U (&u)()) {
-    typename common_type_t<T, U>;
-    typename common_type_t<U, T>;
-    requires Same<common_type_t<U, T>, common_type_t<T, U>>();
-    common_type_t<T, U>(t());
-    common_type_t<T, U>(u());
-    requires CommonReference<add_lvalue_reference_t<common_type_t<T, U>>,
-                             common_reference_t<add_lvalue_reference_t<const T>,
-                                                add_lvalue_reference_t<const U>>>();
- };
+ return
+   Same<common_type_t<T, U>, common_type_t<U, T>>() && 
+   ConvertibleTo<T, common_type_t<T, U>>() &&
+   ConvertibleTo<U, common_type_t<T, U>>() &&
+   CommonReference<add_lvalue_reference_t<const T>,
+                   add_lvalue_reference_t<const U>>() &&
+   CommonReference<add_lvalue_reference_t<common_type_t<T, U>>,
+                   common_reference_t<add_lvalue_reference_t<const T>,
+                                      add_lvalue_reference_t<const U>>>()
}
```

*Note:* this removes the `typename common_type_t...` constraints since they are superfluous. It also fixes a bug where the original `Common` definition was trying to form the types `const T&` and `const U&`. That fails for _cv_ `void`. We would like `Common<void, void>()` to be `true`.

cjdb:


concepts.tex
```diff
@@ -445,16 +440,12 @@
...
-      common_type_t<T, U>(u());
-      requires CommonReference<add_lvalue_reference_t<common_type_t<T, U>>,
-                               common_reference_t<add_lvalue_reference_t<const T>,
-                                                  add_lvalue_reference_t<const U>>>();
-    };
+    Same<common_type_t<U, T>, common_type_t<T, U>>() &&
+    ConvertibleTo<T, common_type_t<T, U>>() &&
+    ConvertibleTo<U, common_type_t<T, U>>() &&
+    CommonReference<add_lvalue_reference_t<common_type_t<T, U>>,
+                     common_reference_t<add_lvalue_reference_t<const T>,
```
CaseyCarter: `common_reference_t` is not aligned with the above `add_lvalue_reference_t`.

concepts.tex
```diff
@@ -403,14 +403,9 @@
...
-    typename common_reference_t<T, U>;
-    typename common_reference_t<U, T>;
-    requires Same<common_reference_t<T, U>,
-                  common_reference_t<U, T>>();
-    common_reference_t<T, U>(t());
-    common_reference_t<T, U>(u());
-  };
+  return Same<common_reference_t<T, U>, common_reference_t<U, T>>() &&
+  ConvertibleTo<T, common_reference_t<T, U>>() &&
+  ConvertibleTo<U, common_reference_t<T, U>>();
```
CaseyCarter: I prefer the formatting in the PR for this definition.

CaseyCarter:
Superseded by #449.

Issue #345, opened 2017-02-28
US 2 (006): 2.1.1: Update ranged-for-loop wording


CaseyCarter:
### Editorial Comment

The wording no longer matches the more thoroughly reviewed form in the C++17 Working Draft.

### Proposed Change

Entirely replace this wording with the proposed wording for C++17.

ericniebler:
LEWG says yes, do it.

CaseyCarter:
Our base document is C++14. I think before we can do this we need to determine - possibly with advisement from CWG - if both versions of the wording are normatively equivalent. I certainly do NOT want to require implementations of C++14 + Ranges to implement any other normative differences to range-based-for other than the relaxation to enable sentinel delimited ranges.

CaseyCarter:
CWG (in telecon) sensibly requests full text of the proposed change to the TS wording.

CaseyCarter:
CWG in telecon approves the wording of the proposed change and is calling it tentatively ready. I am to get a P-number and submit it in the pre-meeting mailing to be moved in Toronto.

ericniebler:
Awesome, thanks Casey!

PR #400, opened 2017-04-10
Integrate C++17 FDIS wording for stmt.ranged with diff marks

CaseyCarter:
Addresses #345 (US-2)

### DIFF
--- a/statements.tex
+++ b/statements.tex
@@ -7,7 +7,7 @@
 \rSec2[stmt.ranged]{The range-based \tcode{for} statement}
 
 \ednote{Modify \cxxref{stmt.ranged}/p1 to allow differently typed begin and end
-iterators, like in C++17.}
+iterators, as in C++17:}
 
 \begin{quote}
 \pnum
@@ -40,8 +40,9 @@ iterators, like in C++17.}
 \terminal{\{}\br
 \>\terminal{auto \&\&__range =} range-init \terminal{;}\br
 \>\terminal{for ( auto __begin =} begin-expr \terminal{,}\br
-\>\>\terminal{ __end =} end-expr \terminal{;}\br
-\>\>\terminal{ __begin != __end; ++__begin ) \{}\br
+\>\terminal{{ }{ }{ }{ }{ }{ }{ }{ }{ }{ }{ }__end =} end-expr \terminal{;}\br
+\>\>\terminal{ __begin != __end;}\br
+\>\>\terminal{ ++__begin ) \{}\br
 \>\>for-range-declaration \terminal{= *__begin;}\br
 \>\>statement\br
 \>\terminal{\}}\br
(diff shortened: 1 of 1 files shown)


statements.tex
```diff
@@ -11,79 +11,96 @@
...
-\begin{bnf}
-\terminal{for} \terminal{(} for-range-declaration \terminal{:} expression \terminal{)} statement
-\end{bnf}
+\begin{ncbnf}
+\terminal{for (} for-range-declaration \terminal{:} \oldoldtxt{expression}\newnewtxt{for-range-initializer} \terminal{)} statement
+\end{ncbnf}
 %
-let \grammarterm{range-init} be equivalent to the expression surrounded by parentheses
+\begin{removedblock}
+\oldoldtxt{let \grammarterm{range-init} be equivalent to the expression surrounded by parentheses}
```
ericniebler: Footnote 89 here ("this ensures that a top-level comma operator cannot be reinterpreted [...]") also needs to be shown as deleted.

statements.tex
```diff
@@ -11,79 +11,96 @@
...
+it is regarded as if it were surrounded by parentheses (so that a comma operator
+cannot be reinterpreted as delimiting two \grammarterm{init-declarator}{s});}
+
+\item
+\tcode{__range}, \tcode{__begin}, and \tcode{__end} are variables defined for
+exposition only, and \oldoldtxt{\tcode{_RangeT} is the type of the
+\grammarterm{}{expression}, and}
+
+\item
+\placeholder{begin-expr} and \placeholder{end-expr} are determined as follows:
```
ericniebler: In C++14, this line does not get its own bullet. (These core guys are picky; let's not give them any reason to complain.)

statements.tex
CaseyCarter: I'll ednote striking the footnote.

statements.tex
```diff
@@ -11,79 +11,96 @@
...
-    @\grammarterm{for-range-declaration}@ = *__begin;
-    @\grammarterm{statement}@
-  }
-}
-\end{codeblock}
+\begin{ncbnftab}
+\terminal{\{}\br
+\>\terminal{auto \&\&__range =} range-init \terminal{;}\br
+\>\terminal{for ( auto __begin =} begin-expr \terminal{,}\br
+\>\>\terminal{ __end =} end-expr \terminal{;}\br
```
CaseyCarter: Yes, I did some battle with the tex sources here and lost: alignment to non-tab boundaries is not easy in this ncbnftab environment. I've since thought of protecting spaces:
```tex
\terminal{{ }{ }{ }{ }{ }__end = } ...
```
which seems to work.

statements.tex
```diff
@@ -11,79 +11,96 @@
...
+
+\begin{itemize}
+\item
+\newnewtxt{if the \grammarterm{for-range-initializer} is an \grammarterm{expression},
+it is regarded as if it were surrounded by parentheses (so that a comma operator
+cannot be reinterpreted as delimiting two \grammarterm{init-declarator}{s});}
+
+\item
+\tcode{__range}, \tcode{__begin}, and \tcode{__end} are variables defined for
+exposition only, and \oldoldtxt{\tcode{_RangeT} is the type of the
```
CaseyCarter: Good catch. I'll omit diff marking this change - I think it would hinder rather than help understand the differences.

statements.tex
CaseyCarter: My thought was that the line break and bullet are purely editorial differences, and that I should minimize the markup to make the overall result easier to review. That said, someone will be comparing the text to 14 and putting the old/new markup in will prevent them wasting a minute or so figure out what's going on here. Will change.

PR #408, opened 2017-05-20
[EDITORIAL] Merge bite-sized iterator operation subclauses into bigge…

### LINKED ISSUE #406: Consider reducing the number of subclauses in [iterators]
In particular, we probably don't need one subclause per operator for many of the iterator types. For `reverse_iterator`, for instance, the current setup results in 3 consecutive pages each with half a dozen subclauses on them.

`cxxdraft-htmlgen` reports that [iterators] has a whopping 194 sections, which is more than half of the number of sections in the entire TS (371).

timsong-cpp:
…r ones.

Fixes #406.

### DIFF
--- a/iterators.tex
+++ b/iterators.tex
@@ -2390,9 +2390,7 @@ namespace std { namespace experimental { namespace ranges { inline namespace v1 
 }}}}
 \end{codeblock}
 
-\rSec3[reverse.iter.ops]{\tcode{reverse_iterator} operations}
-
-\rSec4[reverse.iter.cons]{\tcode{reverse_iterator} constructor}
+\rSec3[reverse.iter.cons]{\tcode{reverse_iterator} constructors and assignment}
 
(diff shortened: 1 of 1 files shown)


timsong-cpp:
This reduces the number of sections in [iterators] by 57, to 137.

ericniebler:
No argument that the extra headings are unnecessary. But that's how the IS does it, and doing it differently will make it harder for implementors and users to cross-reference between the two documents. That's not a strong argument for or against, just sayin'.

timsong-cpp:
The IS seems to be moving in the direction of merging them too (see `istreambuf_iterator`).

CaseyCarter:
> The IS seems to be moving

C++14 is not moving; it's quite stationary. I prefer the organization to mirror that of C++14 as much as possible to ease the learning curve.

When we integrate into the IS, I would *love* to make this change globally in the IS.

PR #412, opened 2017-06-15
[EDITORIAL] Clarify that alg.foreach has stronger requirements than in C++14

### LINKED ISSUE #401: for_each requirements strengthened vs. C++14
`std::for_each` only requires its function argument to be `MoveConstructible`, whereas `ranges::for_each` requires its function argument to be `CopyConstructible`. Is this strengthening of requirements intentional?

We should either relax the requirement to agree with `std::for_each` or add a note to specifically point out that the requirement is stronger than that of C++14.

CaseyCarter:
Fixes #401.

### DIFF
--- a/algorithms.tex
+++ b/algorithms.tex
@@ -1262,6 +1262,12 @@ times.
 \pnum
 \notes
 If \tcode{f} returns a result, the result is ignored.
+
+\pnum
+\enternote The requirements of this algorithm are more strict than those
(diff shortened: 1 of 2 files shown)


ericniebler:
I personally don't feel this is necessary, but I'll merge anyway.

CaseyCarter:
> I personally don't feel this is necessary, but I'll merge anyway.

I would agree if `CopyConstructible` was explicit in the algorithm specification as is `MoveConstructible` in C++14, rather then being hidden inside `IndirectInvocable`.

PR #437, opened 2017-07-17
[EDITORIAL] Correct mispellings of "only if" as "if and only if"

CaseyCarter:


iterators.tex
```diff
@@ -1042,10 +1042,10 @@
 \pnum
 Let \tcode{i} be an object of type \tcode{I}. When both pre- and post-increment
 are valid, \tcode{i} is said to be \techterm{incrementable}. Then
-\tcode{WeaklyIncrementable<I>} is satisfied if and only if
+\tcode{WeaklyIncrementable<I>} is satisfied only if
 
 \begin{itemize}
-\item \tcode{++i} is valid if and only if \tcode{i++} is valid.
+\item The expressions \tcode{++i} and \tcode{i++} have the same domain.
```
ericniebler: 👍

Issue #314, opened 2017-01-27
ConvertibleTo<T&&, U> should say something about the final state of the source object


ericniebler:
`Assignable` and `MoveConstructible` explicitly say that the source object may end up in a moved-from state as a result of the operation. I think `ConvertibleTo` should also.

Proposed resolution
--------------------

Fixed by the proposed resolution of #167.

PR #438, opened 2017-07-17
Apply the PR of #167 from P0740R0

### LINKED ISSUE #167: ConvertibleTo should require both implicit and explicit conversion
Currently, `ConvertibleTo` checks only for _implicit_ convertibility (a-la `is_convertible`). It would be highly surprising for generic code for an _explicit_ conversion to fail, or to succeed but yield a different result.

### Proposed Resolution

(Also fixes #314.)

Replace the contents of [concepts.lib.corelang.convertibleto] with:

> ```c++
> template <class From, class To>
> concept bool ConvertibleTo() {
>   return is_convertible<From, To>::value &&
>     requires (From (&f)()) {
>       static_cast<To>(f());
>     };
> }
> ```
> 
> 1 Let `test` be the invented function:
> ```c++
> To test(From (&f)()) {
>   return f();
> }
> ```
> and let `f` be a function with no arguments and return type `From` such that `f()` is equality preserving. Then `ConvertibleTo<From, To>()` is satisfied if and only if:
>
> > (1.1) - `To` is not an object or reference-to-object type, or `static_cast<To>(f())` is equal to `test(f)`.
> >  (1.2) - `From` is not a reference-to-object type, or
> > > (1.2.1) - If `From` is an rvalue reference to a non `const`-qualified type, the resulting state of the object referenced by `f()` after either above expression is valid but unspecified ([lib.types.movedfrom]).
> > > (1.2.2) - Otherwise, the object referred to by `f()` is not modified by either above expression.
>
>2 There need not be any subsumption relationship between `ConvertibleTo<From, To>()`
>  and `is_convertible<From, To>::value`.

CaseyCarter:
Fixes #167 and #314.

### DIFF
--- a/concepts.tex
+++ b/concepts.tex
@@ -300,14 +300,48 @@ concept bool DerivedFrom = @\seebelow@;
 \indexlibrary{\idxcode{ConvertibleTo}}%
 \begin{itemdecl}
 template <class T, class U>
-concept bool ConvertibleTo = @\seebelow@;
+concept bool ConvertibleTo =
+  is_convertible<From, To>::value &&
+  requires(From (&f)()) {
+    static_cast<To>(f());
+  };
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
-\tcode{ConvertibleTo<T, U>} is satisfied
-if and only if \tcode{is_convertible<T, U>::value} is \tcode{true}.
+Let \tcode{test} be the invented function:
+\begin{codeblock}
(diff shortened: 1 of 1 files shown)


Casey's WIP.md
```diff
@@ -0,0 +1,13 @@
+### P0663 "Ranges TS 'Ready' Issues for Toronto"
```
CaseyCarter: I will delete this tracking file when this round of edits is complete.

Issue #330, opened 2017-02-16
Argument deduction constraints are specified incorrectly


ericniebler:
See #155 for detailed discussion, but basically we've been operating under the faulty assumption that the following:

```c++
requires(const T& t) {  {t} -> Same<const T&>; }
```

was trivially satisfied. It is not. Argument deduction constraints are handled as if the placeholder were made the argument type of an invented function `f`, like `void f(Same<const T&>)`, and then a call were made like `f(t)`. That causes the type of `t` to decay.

This should instead be written as:

```c++
requires(const T& t) {  {t} -> Same<const T&>&&; }
```

That prevents argument decay from happening, which recovers the value category information.

We need to review and amend _all_ the definitions and uses of concepts in the ranges ts.

(In our defense, this slipped through because early concepts implementations didn't do argument deduction constraints, so we invented a workaround, and the workaround was buggy because we didn't understand the model.)

### Proposed Resolution:

Update section "Concept `Assignable`" ([concepts.lib.corelang.assignable]) and "Concept `Destructible`" ([concepts.lib.object.destructible]) as specified in [P0547R1](http://wg21.link/P0547R1).

Update sections "Concept `Boolean`" [concepts.lib.compare.boolean]), "Concept `EqualityComparable`" ([concepts.lib.compare.equalitycomparable]), "Concept `StrictTotallyOrdered`" ([concepts.lib.compare.stricttotallyordered]) as specified in #155. 

Update section "Concept `Readable`" ([iterators.readable]) as follows (also fixes #339):

```diff
template <class I>
concept bool Readable() {
- return Movable<I>() && DefaultConstructible<I>() &&
-   requires(const I& i) {
+ return requires {
    typename value_type_t<I>;
    typename reference_t<I>;
    typename rvalue_reference_t<I>;
-   { *i } -> Same<reference_t<I>>;
-   { ranges::iter_move(i) } -> Same<rvalue_reference_t<I>>;
  } &&
  CommonReference<reference_t<I>, value_type_t<I>&>() &&
  CommonReference<reference_t<I>, rvalue_reference_t<I>>() &&
  CommonReference<rvalue_reference_t<I>, const value_type_t<I>&>();
}
```

Change section "Concept `WeaklyIncrementable`" ([iterators.weaklyincrementable]) as follows:

```diff
template <class I>
concept bool WeaklyIncrementable() {
  return Semiregular<I>() &&
  requires(I i) {
    typename difference_type_t<I>;
[...]

PR #440, opened 2017-07-17
Issue 155

### LINKED ISSUE #155: Comparison concepts and reference types
We have quite a few concepts in the TS that are well-defined over value types - un-cv-qualified non-array object types - whose meaning is unclear for non-value types. We need to determine what their meaning should be. I'll examine `EqualityComparable` as a representative example in hopes that whatever conclusion I reach can be generalized to cover this entire class of concepts.
## `EqualityComparable`

`EqualityComparable<foo>()` has a straight-forward meaning for a value type `foo`:
- `bool(a == b)` iff `a` equals `b` (`==` means "equals")
- `bool(a != b) == !bool(a == b)` (As relations, `!=` is the complement of `==`)
- `a == b` and `a != b` are valid non-modifying equality-preserving expressions if both `a` and `b` are expressions with type `foo` or `const foo` and any value category.

(Aside: The requirement "`bool(a == a)`" is not axiomatic; it's implied by "`bool(a == b)` iff `a` equals `b`". It also seems reasonable to me to require that the definition spaces of the two expressions be equal. Neither of these observations is in scope for this particular discussion.)

What meaning, if any, should `EqualityComparable<const foo>()` have? `EqualityComparable<foo&>()`? `EqualityComparable<volatile foo&&>()`?
## Status Quo

What meaning does the current specification of `EqualityComparable` ascribe to non-value types?
- The very first sentence in the prose, "Let `a` and `b` be objects of type `T`," doesn't have a well-defined meaning when `T` is a non-object type.
- If `T` is a function or array type, the parameters of the requires clause are effectively of type `const decay_t<T>`; parameters are decayed just as function parameters are. As a result, `EqualityComparable<int(double)>() == EqualityComparable<int(*)(double)>()` and `EqualityComparable<int[42]>() == EqualityComparable<int*>()`, neither of which seems sensible to me since we cannot directly compare functions nor arrays for equality in C++. Even more bizarre is the cross-type case, where `EqualityComparable<int[13],int[42]>() == EqualityComparable<int*,int*>()`.
- If `T` is a reference type, `const T` is the same type as `T`, so the required expressions must be valid for lvalues of type `remove_reference_t<T>`. If `remove_reference_t<T>` is not `const`, the expressions may modify the operands and they do not implicitly require expression variants.
- cv-qualified non-reference `T` actually seems meaningful. "`a` equals `b`" is defined for values, and object types that differ in their cv-qualifiers have the same value type.
  - `const const T` is the same type as `const T` so the required expressions are exactly equivalent to the expressions required by unqualified `T` when `T` is `const`-qualified. This would not be the case for a concept with expressions that are intended to allow modification.  Given a hypothetical concept `C<T>` that requires an expression `f(t)` that may modify `t`, `C<const T>` requires `f(ct)` to be non-modifying and have implicit expression variants that also must be non-modifying. It's subtle that a template with associated constraints that include both `C<foo>` and `C<const foo>` actually forbids `f` to modify a mutable lvalue `foo`.
  - The case for `volatile` is similar: the explicitly required expressions operate on lvalues of type `const volatile T`, so implicit expression variants are required for all value categories of expressions with type `const volatile T` or `volatile T`, but notably not types `T` or `const T`. This could be addressed by extending the wording for implicit expression variants to generally cover all less-cv-qualified types.
    
    The same considerations apply for `C<volatile foo>` and `C<const volatile foo>` as noted above for `C<foo>` and `C<const foo>`.
    
    Note that the expression stability requirement in [concepts.lib.general.equality]/3 effectively disables the meaning of `volatile` in equality-preserving expressions: an external process or action may change the value of a `volatile` object, but it's forbidden to do so when the library may observe it. This distinction is subtle and may cause confusion if value-semantic concepts admit `volatile` types.
## Functions & Arrays

The consequences of allowing function and array types are both bizarre and subtle, I believe the entire class of value-semantic concepts should categorically reject them.
## References

The effects of allowing reference types are also quite subtle. The vast majority of non-language lawyers don't know how cv-qualifiers apply to reference types and won't have a chance of intuiting what `EqualityComparable<Foo&&>` means. Admitting reference types leads to a particular logic error I call "type/reference confusion": it's not clear to the reader or writer of code whether the intent is that the concept requirements apply to the type of the _reference_, or apply to the type of the _referent_. Proxy reference types exacerbate this problem in that they aren't easily distinguishable as references. If users become accustomed to `C<Foo&>` having the same meaning as `C<Foo>`, they will be caught unawares when e.g. `C<reference_wrapper<Foo>>` does not.

Until we have more widespread user experience with the TS I think it's best that the value-semantic class of concepts not admit reference types. What's not clear to me is whether the concepts should reject reference types syntactically, or if a program that applies such a concept to a reference type should be ill-formed. I've used both techniques for various concepts in the implementation of the TS at different times. My experience - which can hardly be considered statistically representative - is that hard errors usually occur closer to the source of the problem, and that silent rejections result in hard to find bugs buried beneath layers of templates and concepts.
## cv-qualifiers

I'm least certain about how to approach cv-qualifiers. I can see a path that will give them meaning - much as we did with the object concepts - although there are still some subtleties. How we choose to handle cv-qualifiers has ramifications for the rest of the design. An `Iterator` cannot currently have a `volatile`-qualified reference type, for example, and allowing that would require changes throughout the algorithm specifications. Supporting `volatile`-qualified types in the concepts provides little value if the rest of the library doesn't support them, and changing the entire TS to support a freakish corner case is out of the question.

`const`-qualified types are easily supportable with reasonable semantics, and less marginal than `volatile` types. I'm fairly certain the existing wording of the library concepts works with `const` types and would require no changes. Forbidding reference types and volatile qualifiers while allowing `const` does seem like a special case to me, and therefore undesirable for consistency's sake.
## Conclusions

If I had to choose an approach right now I would keep the meaning of the concepts as sharply defined as possible by rejecting non-object types, array types, and cv-qualified types syntactically, while making reference types ill-formed. Users must then learn that satisfaction of `C<remove_const_t<T>>()` implies that all of the required non-modifying expressions are valid for `T`, but the required maybe-modifying expressions are only valid if `T` is not `const`. Forcing users to deal with cv-qualifiers and reference types explicitly makes it harder to write generic code, but simultaneously makes it easier to write generic code with fewer errors.

Proposed Resolution
-------------------

Change the definition of the `Booolean` concept ([concepts.lib.compare.boolean]/p1) as follows (includes the resolution for #330):

```diff
template <class B>
concept bool Boolean() {
-  return MoveConstructible<B>() && // (see 4.5.4)
-    requires(const B b1, const B b2, const bool a) {
-      bool(b1);
-      { b1 } -> bool;
-      bool(!b1);
-      { !b1 } -> bool;
-      { b1 && b2 } -> Same<bool>;
-      { b1 && a } -> Same<bool>;
-      { a && b1 } -> Same<bool>;
-      { b1 || b2 } -> Same<bool>;
-      { b1 || a } -> Same<bool>;
-      { a || b1 } -> Same<bool>;
-      { b1 == b2 } -> bool;
-      { b1 != b2 } -> bool;
-      { b1 == a } -> bool;
-      { a == b1 } -> bool;
-      { b1 != a } -> bool;
-      { a != b1 } -> bool;
+  return Movable<decay_t<B>>() && // (see \ref{concepts.lib.object.movable})
+    requires(const remove_reference_t<B>& b1,
+             const remove_reference_t<B>& b2, const bool a) {
+      { b1 } -> ConvertibleTo<bool>&&;
+      { !b1 } -> ConvertibleTo<bool>&&;
+      { b1 && a } ->  Same<bool>&&;
+      { b1 || a } ->  Same<bool>&&;
+      { b1 && b2 } -> Same<bool>&&;
+      { a && b2  } -> Same<bool>&&;
+      { b1 || b2 } -> Same<bool>&&;
+      { a || b2  } -> Same<bool>&&;
+      { b1 == b2 } -> ConvertibleTo<bool>&&;
+      { b1 == a  } -> ConvertibleTo<bool>&&;
+      { a == b2  } -> ConvertibleTo<bool>&&;
+      { b1 != b2 } -> ConvertibleTo<bool>&&;
+      { b1 != a  } -> ConvertibleTo<bool>&&;
+      { a != b2  } -> ConvertibleTo<bool>&&;
    };
 }
```

Change [concepts.lib.compare.boolean]/p2 as follows (depends on the resolution of #167):

```diff
-2 Given values b1 and b2 of type B, then Boolean<B>() is satisfied if and only if
+2 Given const lvalues b1 and b2 of type remove_reference_t<B>, then
+  Boolean<B>() is satisfied if and only if
-(2.1) — bool(b1) == [](bool x) { return x; }(b1).
 (2.2) — bool(b1) == !bool(!b1).
 (2.3) — (b1 && b2), (b1 && bool(b2)), and (bool(b1) && b2) are all equal to 
         (bool(b1) && bool(b2)), and have the same short-circuit evaluation.
 (2.4) — (b1 || b2), (b1 || bool(b2)), and (bool(b1) || b2) are all equal to
         (bool(b1) || bool(b2)), and have the same short-circuit evaluation.
 (2.5) — bool(b1 == b2), bool(b1 == bool(b2)), and bool(bool(b1) == b2) are
         all equal to (bool(b1) == bool(b2)).
 (2.6) — bool(b1 != b2), bool(b1 != bool(b2)), and bool(bool(b1) != b2) are
         all equal to (bool(b1) != bool(b2)).
```

Change concept `WeaklyEqualityComparable` ([concepts.lib.compare.equalitycomparable]) as follows (includes the resolution for #330):

```diff
 template <class T, class U>
 concept bool WeaklyEqualityComparable() {
-  return requires(const T& t, const U& u) {
-    { t == u } -> Boolean;
-    { u == t } -> Boolean;
-    { t != u } -> Boolean;
-    { u != t } -> Boolean;
+  return requires(const remove_reference_t<T>& t,
+                  const remove_reference_t<U>& u) {
+    { t == u } -> Boolean&&;
+    { t != u } -> Boolean&&;
+    { u == t } -> Boolean&&;
+    { u != t } -> Boolean&&;
   };
 }
```

Change [concepts.lib.compare.equalitycomparable]/p1 as follows:

```diff
-1 Let t and u be objects of types T and U.
+1 Let t and u be const lvalues of types remove_reference_t<T> and remove_reference_t<U>.
   WeaklyEqualityComparable<T, U>() is satisfied if and only if:
 (1.1) — t == u, u == t, t != u, and u != t have the same domain.
 (1.2) — bool(u == t) == bool(t == u).
 (1.3) — bool(t != u) == !bool(t == u).
```

Change cross-type concept `EqualityComparable` ([concepts.lib.compare.equalitycomparable]) as follows (includes the resolution for #330):

```diff
 template <class T, class U>
 concept bool EqualityComparable() {
-  return CommonReference<const T&, const U&>() &&
+  return
     EqualityComparable<T>() &&
     EqualityComparable<U>() &&
-    EqualityComparable<
-        remove_cv_t<remove_reference_t<common_reference_t<const T&, const U&>>>>() &&
+    CommonReference<
+      const remove_reference_t<T>&,
+      const remove_reference_t<U>&>() &&
+    EqualityComparable<
+      common_reference_t<
+        const remove_reference_t<T>&,
+        const remove_reference_t<U>&>>() &&
     WeaklyEqualityComparable<T, U>();
 }
```

Change [concepts.lib.compare.equalitycomparable]/p4 as follows:

```diff
-4 Let a be an object of type T, b be an object of type U, and C be
-  common_reference_t<const T&, const U&>.
+4 Let a be a const lvalue of type remove_reference_t<T>, b be a
+  const lvalue of type remove_reference_t<U>, and C be
+  common_reference_t<const remove_reference_t<T>&,
+  const remove_reference_t<U>&>.
   Then EqualityComparable<T, U>() is satisfied if and only if:
 (4.1) — bool(a == b) == bool(C(a) == C(b)).
```

Change concept `StrictTotallyOrdered` ([concepts.lib.compare.stricttotallyordered]) as follows (includes the resolution for #330):

```diff
 template <class T>
 concept bool StrictTotallyOrdered() {
   return EqualityComparable<T>() &&
-    requires(const T a, const T b) {
+    requires(const remove_reference_t<T>& t,
+             const remove_reference_t<U>& u) {
-      { a < b } -> Boolean;
-      { a > b } -> Boolean;
-      { a <= b } -> Boolean;
-      { a >= b } -> Boolean;
+      { a < b } -> Boolean&&;
+      { a > b } -> Boolean&&;
+      { a <= b } -> Boolean&&;
+      { a >= b } -> Boolean&&;
     };
 }
```

Change [concepts.lib.compare.stricttotallyordered]/p1 to be:

```diff
-1 Let a, b, and c be objects of type T.
+1 Let a, b, and c be const lvalues of type remove_reference_t<T>.
 Then StrictTotallyOrdered<T>() is satisfied if and only if
 (1.1) — Exactly one of bool(a < b), bool(b < a), or bool(a == b) is true.
 (1.2) — If bool(a < b) and bool(b < c), then bool(a < c).
 (1.3) — bool(a > b) == bool(b < a).
 (1.4) — bool(a <= b) == !bool(b < a).
 (1.5) — bool(a >= b) == !bool(a < b).
```

Change cross-type concept `StrictTotallyOrdered` ([concepts.lib.compare.stricttotallyordered]) as follows (includes the resolution for #330):

```diff
 template <class T, class U>
 concept bool StrictTotallyOrdered() {
-  return CommonReference<const T&, const U&>() &&
+  return
     StrictTotallyOrdered<T>() &&
     StrictTotallyOrdered<U>() &&
-    StrictTotallyOrdered<
-      remove_cv_t<remove_reference_t<common_reference_t<const T&, const U&>>>>() &&
+    CommonReference<
+      const remove_reference_t<T>&,
+      const remove_reference_t<U>&>() &&
+    StrictTotallyOrdered<
+      common_reference_t<
+        const remove_reference_t<T>&,
+        const remove_reference_t<U>&>>() &&
     EqualityComparable<T, U>() &&
-    requires(const T t, const U u) {
+    requires(const remove_reference_t<T>& t,
+             const remove_reference_t<U>& u) {
-      { t < u } -> Boolean;
-      { t > u } -> Boolean;
-      { t <= u } -> Boolean;
-      { t >= u } -> Boolean;
-      { u < t } -> Boolean;
-      { u > t } -> Boolean;
-      { u <= t } -> Boolean;
-      { u >= t } -> Boolean;
+      { t < u } -> Boolean&&;
+      { t > u } -> Boolean&&;
+      { t <= u } -> Boolean&&;
+      { t >= u } -> Boolean&&;
+      { u < t } -> Boolean&&;
+      { u > t } -> Boolean&&;
+      { u <= t } -> Boolean&&;
+      { u >= t } -> Boolean&&;
     };
 }
```

Change [concepts.lib.compare.stricttotallyordered]/p2 as follows:

```diff
-2 Let t be an object of type T, u be an object of type U,
+2 Let t be a const lvalue of type remove_reference_t<T>,
+  u be a const lvalue of type remove_reference_t<U>,
-  and C be common_reference_t<const T&, const U&>.
+  and C be common_reference_t<const remove_reference_t<T>&,
+  const remove_reference_t<U>&>.
   Then StrictTotallyOrdered<T, U>() is satisfied if and only if
 (2.1) — bool(t < u) == bool(C(t) < C(u)).
 (2.2) — bool(t > u) == bool(C(t) > C(u)).
 (2.3) — bool(t <= u) == bool(C(t) <= C(u)).
 (2.4) — bool(t >= u) == bool(C(t) >= C(u)).
 (2.5) — bool(u < t) == bool(C(u) < C(t)).
 (2.6) — bool(u > t) == bool(C(u) > C(t)).
 (2.7) — bool(u <= t) == bool(C(u) <= C(t)).
 (2.8) — bool(u >= t) == bool(C(u) >= C(t)).
```

Change section "Concept `Relation`" ([concepts.lib.callable.relation]) as follows:

```diff
template <class R, class T, class U>
concept bool Relation() {
  return Relation<R, T>() &&
    Relation<R, U>() &&
-    CommonReference<const T&, const U&>() &&
-    Relation<R,
-      common_reference_t<const T&, const U&>>() &&
+    CommonReference<
+      const remove_reference_t<T>&,
+      const remove_reference_t<U>&>() &&
+    Relation<R,
+      common_reference_t<
+        const remove_reference_t<T>&,
+        const remove_reference_t<U>&>>() &&
    Predicate<R, T, U>() &&
    Predicate<R, U, T>();
}

-1 Let r be any object of type R,
+1 Let r be an expression such that decltype((r)) is R,
-  a be any object of type T,
+  a be an expression such that decltype((a)) is T,
- b be any object of type U,
+  b be an expression such that decltype((b)) is U,
-  and C be common_reference_t<const T&, const U&>.
+  and C be common_reference_t<const remove_reference_t<T>&,
+  const remove_reference_t<U>&>.
  Then Relation<R, T, U>() is satisfied if and only if
(1.1) — bool(r(a, b)) == bool(r(C(a), C(b))).
(1.2) — bool(r(b, a)) == bool(r(C(b), C(a))).
```

Change "Concept `Swappable`" ([concepts.lib.corelang.swappable]) as follows:

```diff
template <class T>
concept bool Swappable() {
  return requires(T&& a, T&& b) {
    ranges::swap(std::forward<T>(a), std::forward<T>(b));
  };
}

template <class T, class U>
concept bool Swappable() {
  return Swappable<T>() &&
  Swappable<U>() &&
- CommonReference<const T&, const U&>() &&
+ CommonReference<
+   const remove_reference_t<T>&,
+   const remove_reference_t<U>&>() &&
  requires(T&& t, U&& u) {
    ranges::swap(std::forward<T>(t), std::forward<U>(u));
    ranges::swap(std::forward<U>(u), std::forward<T>(t));
  };
}
```

CaseyCarter:
Apply P0663 PR for Issue 155: Comparison concepts and reference types

Fixes #155; partially addresses #330. Depends on #167.

### DIFF
--- a/concepts.tex
+++ b/concepts.tex
@@ -316,7 +316,7 @@ To test(From (&f)()) {
 }
 \end{codeblock}
 and let \tcode{f} be a function with no arguments and return type \tcode{From}
-such that \tcode{f()} is equality preserving. Then
+such that \tcode{f()} is equality preserving.
 \tcode{ConvertibleTo<From, To>} is satisfied only if:
 
 \begin{itemize}
@@ -485,7 +485,7 @@ type, let \tcode{v} be an lvalue of type \tcode{R};
 otherwise, let \tcode{v} be an rvalue of type \tcode{R}.
 Let \tcode{uu} be a distinct object of type \tcode{R} such that
 \tcode{uu} is equal to \tcode{v}.
-Then \tcode{Assignable<T, U>} is satisfied only if
+\tcode{Assignable<T, U>} is satisfied only if
 
 \begin{itemize}
 \item \tcode{addressof(t = v) == addressof(t)}.
@@ -510,7 +510,9 @@ concept bool Swappable =
 
 template <class T, class U>
 concept bool SwappableWith =
-  CommonReference<const T&, const U&> &&
+  CommonReference<
+    const remove_reference_t<T>&,
+    const remove_reference_t<U>&>() &&
   requires(T&& t, U&& u) {
     ranges::swap(std::forward<T>(t), std::forward<T>(t));
     ranges::swap(std::forward<U>(u), std::forward<U>(u));
@@ -599,33 +601,31 @@ The \tcode{Boolean} concept specifies the requirements on a type that is usable 
 \begin{itemdecl}
 template <class B>
 concept bool Boolean =
-  MoveConstructible<B> && // (see \ref{concepts.lib.object.moveconstructible})
-  requires(const B b1, const B b2, const bool a) {
-    bool(b1);
-    { b1 } -> bool;
-    bool(!b1);
-    { !b1 } -> bool;
-    { b1 && b2 } -> Same<bool>;
-    { b1 && a } -> Same<bool>;
-    { a && b1 } -> Same<bool>;
-    { b1 || b2 } -> Same<bool>;
-    { b1 || a } -> Same<bool>;
-    { a || b1 } -> Same<bool>;
-    { b1 == b2 } -> bool;
-    { b1 != b2 } -> bool;
-    { b1 == a } -> bool;
-    { a == b1 } -> bool;
-    { b1 != a } -> bool;
-    { a != b1 } -> bool;
+  Movable<decay_t<B>> && // (see \ref{concepts.lib.object.movable})
+  requires(const remove_reference_t<B>& b1,
+           const remove_reference_t<B>& b2, const bool a) {
+    { b1 }       -> ConvertibleTo<bool>&&;
+    { !b1 }      -> ConvertibleTo<bool>&&;
+    { b1 && a }  -> Same<bool>&&;
+    { b1 || a }  -> Same<bool>&&;
+    { b1 && b2 } -> Same<bool>&&;
+    { a && b2  } -> Same<bool>&&;
+    { b1 || b2 } -> Same<bool>&&;
+    { a || b2  } -> Same<bool>&&;
+    { b1 == b2 } -> ConvertibleTo<bool>&&;
+    { b1 == a  } -> ConvertibleTo<bool>&&;
+    { a == b2  } -> ConvertibleTo<bool>&&;
+    { b1 != b2 } -> ConvertibleTo<bool>&&;
+    { b1 != a  } -> ConvertibleTo<bool>&&;
+    { a != b2  } -> ConvertibleTo<bool>&&;
   };
 \end{itemdecl}
 
 \pnum
-Given values \tcode{b1} and \tcode{b2} of type \tcode{B}, then
-\tcode{Boolean<B>} is satisfied only if
+Given \tcode{const} lvalues \tcode{b1} and \tcode{b2} of type
+\tcode{remove_reference_t<B>}, then \tcode{Boolean<B>} is satisfied only if
 
 \begin{itemize}
-\item \tcode{bool(b1) == [](bool x) \{ return x; \}(b1)}.
 \item \tcode{bool(b1) == !bool(!b1)}.
 \item \tcode{(b1 \&\& b2)}, \tcode{(b1 \&\& bool(b2))}, and
       \tcode{(bool(b1) \&\& b2)} are all equal to
@@ -650,18 +650,20 @@ not \tcode{Boolean} types.\exitexample
 \begin{itemdecl}
 template <class T, class U>
 concept bool WeaklyEqualityComparableWith =
-  requires(const T& t, const U& u) {
-    { t == u } -> Boolean;
-    { u == t } -> Boolean;
-    { t != u } -> Boolean;
-    { u != t } -> Boolean;
+  requires(const remove_reference_t<T>& t,
+           const remove_reference_t<U>& u) {
+    { t == u } -> Boolean&&;
+    { t != u } -> Boolean&&;
+    { u == t } -> Boolean&&;
+    { u != t } -> Boolean&&;
   };
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
-Let \tcode{t} and \tcode{u} be objects of types \tcode{T} and \tcode{U} respectively.
-\tcode{WeaklyEqualityComparableWith<T, U>} is satisfied only if:
+Let \tcode{t} and \tcode{u} be \tcode{const} lvalues of types
+\tcode{remove_reference_t<T>} and \tcode{remove_reference_t<U>} respectively.
+\tcode{Weakly\-Equality\-Comparable\-With<T, U>} is satisfied only if:
 \begin{itemize}
 \item \tcode{t == u}, \tcode{u == t}, \tcode{t != u}, and \tcode{u != t}
       have the same domain.
@@ -695,19 +697,24 @@ implies that \tcode{==} is reflexive, transitive, and symmetric.\exitnote
 \begin{itemdecl}
 template <class T, class U>
 concept bool EqualityComparableWith =
-  CommonReference<const T&, const U&> &&
   EqualityComparable<T> &&
   EqualityComparable<U> &&
+  CommonReference<
+    const remove_reference_t<T>&,
+    const remove_reference_t<U>&> &&
   EqualityComparable<
-    remove_cv_t<remove_reference_t<common_reference_t<const T&, const U&>>>> &&
+    common_reference_t<
+      const remove_reference_t<T>&,
+      const remove_reference_t<U>&>> &&
   WeaklyEqualityComparableWith<T, U>;
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
-Let \tcode{t} be an object of type \tcode{T}, \tcode{u} be an object of type \tcode{U}, and \tcode{C} be
-\tcode{common_reference_t<const T\&, const U\&>}.
-Then \tcode{EqualityComparableWith<T, U>} is satisfied only if:
+Let \tcode{t} be a \tcode{const} lvalue of type \tcode{remove_reference_t<T>}, \tcode{u} be a
+\tcode{const} lvalue of type \tcode{remove_reference_t<U>}, and \tcode{C} be
+\tcode{common_reference_t<const remove_reference_t<T>\&, const remove_reference_t<U>\&>}.
+\tcode{EqualityComparableWith<T, U>} is satisfied only if:
 
 \begin{itemize}
 \item \tcode{bool(t == u) == bool(C(t) == C(u))}.
@@ -721,18 +728,19 @@ Then \tcode{EqualityComparableWith<T, U>} is satisfied only if:
 template <class T>
 concept bool StrictTotallyOrdered =
   EqualityComparable<T> &&
-  requires(const T a, const T b) {
-    { a < b } -> Boolean;
-    { a > b } -> Boolean;
-    { a <= b } -> Boolean;
-    { a >= b } -> Boolean;
+  requires(const remove_reference_t<T>& a,
+           const remove_reference_t<T>& b) {
+    { a < b }  -> Boolean&&;
+    { a > b }  -> Boolean&&;
+    { a <= b } -> Boolean&&;
+    { a >= b } -> Boolean&&;
   };
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
-Let \tcode{a}, \tcode{b}, and \tcode{c} be objects of type \tcode{T}.
-Then \tcode{StrictTotallyOrdered<T>} is satisfied only if
+Let \tcode{a}, \tcode{b}, and \tcode{c} be \tcode{const} lvalues of type \tcode{remove_reference_t<T>}.
+\tcode{StrictTotallyOrdered<T>} is satisfied only if
 
 \begin{itemize}
 \item Exactly one of \tcode{bool(a < b)}, \tcode{bool(a > b)}, or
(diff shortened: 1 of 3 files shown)


ericniebler:
Please fix merge conflicts.

CaseyCarter:
Done. I also changed "Let /*  ... */. Then ConceptFoo<T, U> is satisfied only if" to simply "Let  /* ... */. ConceptFoo<T, U> is satisfied only if."

concepts.tex
```diff
@@ -700,19 +702,25 @@
...
+  CommonReference<
+    const remove_reference_t<T>&,
+    const remove_reference_t<U>&> &&
   EqualityComparable<
-    remove_cv_t<remove_reference_t<common_reference_t<const T&, const U&>>>> &&
+    common_reference_t<
+      const remove_reference_t<T>&,
+      const remove_reference_t<U>&>> &&
   WeaklyEqualityComparableWith<T, U>;
+}
```
ericniebler: 'Sup with the trailing semi?

concepts.tex
```diff
@@ -726,18 +734,19 @@
...
-    { a < b } -> Boolean;
-    { a > b } -> Boolean;
-    { a <= b } -> Boolean;
-    { a >= b } -> Boolean;
+  requires(const remove_reference_t<T>& t,
+           const remove_reference_t<U>& u) {
+    { a < b }  -> Boolean&&;
+    { a > b }  -> Boolean&&;
+    { a <= b } -> Boolean&&;
+    { a >= b } -> Boolean&&;
```
ericniebler: You need to change `a` and `b` here to `t` and `u`.

concepts.tex
CaseyCarter: The semi is correct; the `}` is garbage  here.

concepts.tex
CaseyCarter: WTF is `U`?!? Did I apply the cross-type changes to the single-type concept?

CaseyCarter: > Did I apply the cross-type changes to the single-type concept?

Ugh, no - that's what the PR actually says. I think we need to apply some "editorial" fixes here,

CaseyCarter: Specifically, this concept should read:
```c++
concept bool StrictTotallyOrdered =
  EqualityComparable<T> &&
  requires(const remove_reference_t<T>& a,
           const remove_reference_t<T>& b) {
    { a < b }  -> Boolean&&;
    { a > b }  -> Boolean&&;
    { a <= b } -> Boolean&&;
    { a >= b } -> Boolean&&;
  };
```
Do you agree with me that making this correction falls within the limits of air-quotes "editorial"?

ericniebler: Egad. Yes, I agree it's editorial, or at least DWIM.

concepts.tex
ericniebler: Er, that's what I meant, but you figured it out.

CaseyCarter:
Corrections applied.

PR #453, opened 2017-07-18
Apply P0740R0 PR  for 255: DerivedFrom should be “publicly and unambi…

### LINKED ISSUE #255: DerivedFrom should be publicly and unambiguously
### Proposed Resolution

Change [concepts.lib.corelang.derived] as follows:

```diff
 template <class T, class U>
 concept bool DerivedFrom() {
-  return see below ;
+  return is_base_of<U, T>::value &&
+    is_convertible<remove_cv_t<T>*, remove_cv_t<U>*>::value; // see below
 }

-1 DerivedFrom<T, U>() is satisfied if and only if is_base_of<U, T>::value is
-  true.
+1 There need not be a subsumption relationship between DerivedFrom<T, U>() and either
+  is_base_of<U, T>::value or is_convertible<remove_cv_t<T>*, remove_cv_t<U>*>::value.
+
+2 [Note: DerivedFrom<T, U>() is satisfied if and only if T is publicly and unambiguously
+  derived from U, or T and U are the same class type ignoring cv qualifiers.-end note]
```

CaseyCarter:
…guously”

Fixes  #255.

### DIFF
--- a/concepts.tex
+++ b/concepts.tex
@@ -286,13 +286,21 @@ only if \tcode{T} and \tcode{U} denote the same type.
 \indexlibrary{\idxcode{DerivedFrom}}%
 \begin{itemdecl}
 template <class T, class U>
-concept bool DerivedFrom = @\seebelow@;
+concept bool DerivedFrom =
+  is_base_of<U, T>::value &&
+  is_convertible<remove_cv_t<T>*, remove_cv_t<U>*>::value; // \seebelow
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
-\tcode{DerivedFrom<T, U>} is satisfied if and only if
-\tcode{is_base_of<U, T>::value} is \tcode{true}.
+There need not be any subsumption relationship between \tcode{DerivedFrom<T, U>}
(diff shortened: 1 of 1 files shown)


concepts.tex
```diff
@@ -286,13 +286,21 @@
...
-\tcode{DerivedFrom<T, U>} is satisfied if and only if
-\tcode{is_base_of<U, T>::value} is \tcode{true}.
+There need not be any subsumption relationship between \tcode{DerivedFrom<T, U>}
+and either \tcode{is_base_of<U, T>::value} or
+\tcode{is_convertible<remove_cv_t<T>*, remove_cv_t<U>*>::value}.
+
+\pnum
+\enternote \tcode{DerivedFrom<T, U>} is satisfied if and only if \tcode{T} is
+publicly and unambiguously derived from \tcode{U}, or \tcode{T} and \tcode{U}
+are the same class type ignoring \cv-qualifiers. \exitnote
```
ericniebler: "If an only if" seems correct here to me. Just double-checking that you agree, and that this is intentional.

CaseyCarter: Yes, I do agree, and it is intentional.

PR #461, opened 2017-07-18
Apply P0663R0 fix for 354: JP 1 (015): 6.9.2.2/1: Range doesn’t requi…

### LINKED ISSUE #354: JP 1 (015): 6.9.2.2/1: Range doesn't require begin
### Editorial Comment

`ranges::begin(t)` is missing in “requires” for `Range`.

### Proposed Change

Add `ranges::begin(t);` to “requires”.

CaseyCarter:
…re begin

Fixes #354.

### DIFF
--- a/ranges-lib.tex
+++ b/ranges-lib.tex
@@ -528,6 +528,7 @@ iteration over its elements by providing a \tcode{begin} iterator and an
 template <class T>
 concept bool Range =
   requires(T&& t) {
+    ranges::begin(t); // not necessarily equality-preserving (see below)
     ranges::end(t);
   };
 \end{itemdecl}


ranges-lib.tex
```diff
@@ -528,6 +528,7 @@
 template <class T>
 concept bool Range =
   requires(T&& t) {
+    ranges::begin(t); // not necessarily equality-preserving (see below)
     ranges::end(t);
```
ericniebler: Remind me why there is no such comment about `ranges::end`.

CaseyCarter: `end` is *always* equality-preserving, even for single-pass ranges.

CaseyCarter:
Github is killing me with 500 errors.

`ranges::end` needs no such comment because we always require it to be
equality-preserving. (If a single-pass range is a bounded range, you had
better not increment the end iterator.)

Issue #172, opened 2016-04-18
tagged<Base...> should be implicitly constructible from Base


CaseyCarter:
`tagged<Base, Tags...>` needs to be implicitly constructible from `Base&&` when `MoveConstructible<Base>()`, and from `Base const&` when `CopyConstructible<Base>()`. (To avoid the issue in ericniebler/range-v3#324.)

Proposed Resolution
---------------------

Change the class synopsis of `tagged` in [taggedtup.tagged] as follows:

```diff
 template <class Base, TagSpecifier... Tags>
   requires sizeof...(Tags) <= tuple_size<Base>::value
 struct tagged :
   Base, TAGGET (tagged<Base, Tags...>, Ti, i)... { // see below
   using Base::Base;
   tagged() = default;
   tagged(tagged&&) = default;
   tagged(const tagged&) = default;
+  tagged(Base&&) noexcept(see below)
+    requires MoveConstructible<Base>();
+  tagged(const Base&) noexcept(see below)
+    requires CopyConstructible<Base>();
   tagged &operator=(tagged&&) = default;
   tagged &operator=(const tagged&) = default;
   [...]
```

After [taggedtup.tagged]/9, add the following:

> ```
> tagged(Base &&that) noexcept(see below )
>   requires MoveConstructible<Base>();
> ```
> 10 _Remarks:_ The expression in the `noexcept` is equivalent to:
> ```
> is_nothrow_move_constructible<Base>::value
> ```
> 11 Effects: Initializes `Base` with `std::move(that)`.
> ```
>
> tagged(const Base &that) noexcept(see below )
>   requires CopyConstructible<Base>();
> ```
> 12 _Remarks:_ The expression in the `noexcept` is equivalent to:
> ```
> is_nothrow_copy_constructible<Base>::value
> ```
> 13 Effects: Initializes `Base` with `that`.

CaseyCarter:
`get` needs to be a customization point in order to tag non-standard types.

`tagged` needs to inject the tagged getter types "between" itself and `Base` in the inheritance hierarchy, so that tagged getter names hide (instead of ambiguate) members with the same name in `Base`. i.e., if we envision the current design as "fan inheritance":

<pre>
                Tagged
                 ^  ^
        collect--/  \---Base
        ^     ^
getter1-/ ... \-getterN
</pre>

we want to change the structure to "chain inheritance":

<pre>
Tagged
^
\-getter1
  ^
  \-getter2
    ^
   ...
    \-getterN
[...]

PR #470, opened 2017-07-19
Apply P0579R1: Constexpr for <experimental/ranges/iterator>

### LINKED ISSUE #256: Add constexpr to advance, distance, next, and prev
Also common_iterator(?).

CaseyCarter:
Fixes #256.

### DIFF
--- a/iterators.tex
+++ b/iterators.tex
@@ -219,23 +219,23 @@ namespace std { namespace experimental { namespace ranges { inline namespace v1 
 
   // \ref{iterator.operations}, iterator operations:
   template <Iterator I>
-    void advance(I& i, difference_type_t<I> n);
+    constexpr void advance(I& i, difference_type_t<I> n);
   template <Iterator I, Sentinel<I> S>
-    void advance(I& i, S bound);
+    constexpr void advance(I& i, S bound);
   template <Iterator I, Sentinel<I> S>
-    difference_type_t<I> advance(I& i, difference_type_t<I> n, S bound);
+    constexpr difference_type_t<I> advance(I& i, difference_type_t<I> n, S bound);
   template <Iterator I, Sentinel<I> S>
-    difference_type_t<I> distance(I first, S last);
+    constexpr difference_type_t<I> distance(I first, S last);
   template <Iterator I>
-    I next(I x, difference_type_t<I> n = 1);
+    constexpr I next(I x, difference_type_t<I> n = 1);
   template <Iterator I, Sentinel<I> S>
-    I next(I x, S bound);
+    constexpr I next(I x, S bound);
   template <Iterator I, Sentinel<I> S>
-    I next(I x, difference_type_t<I> n, S bound);
+    constexpr I next(I x, difference_type_t<I> n, S bound);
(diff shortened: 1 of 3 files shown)


utilities.tex
```diff
@@ -785,28 +792,28 @@
     tagged(const tagged&) = default;
     tagged &operator=(tagged&&) = default;
     tagged &operator=(const tagged&) = default;
-    tagged(Base&&) noexcept(@\seebelow@)
+    constexpr tagged(Base&&) noexcept(@\seebelow@)
       requires MoveConstructible<Base>;
-    tagged(const Base&) noexcept(@\seebelow@)
+    constexpr tagged(const Base&) noexcept(@\seebelow@)
```
ericniebler: You thought you could sneak these by me, didn't you? ;-) P0579 doesn't specify these two to be `constexpr`. These constructors came from the resolution of #172, which also doesn't specify them to be `constexpr`. If we're playing by the book, we should leave the `constexpr` off here and file an issue. I'm content to look the other way if you wanted to leave them in, though.

iterators.tex
```diff
@@ -244,48 +244,47 @@
...
       const reverse_iterator<I1>& x,
       const reverse_iterator<I2>& y);
   template <RandomAccessIterator I>
-    reverse_iterator<I>
-      operator+(
-    difference_type_t<I> n,
-    const reverse_iterator<I>& x);
+    constexpr reverse_iterator<I> operator+(
+      difference_type_t<I> n,
+      const reverse_iterator<I>& x);
```
ericniebler: Wow that was some phenomenally bad formatting right there.

iterators.tex
```diff
@@ -3723,38 +3717,38 @@
...
     constexpr move_sentinel();
-    explicit move_sentinel(S s);
-    move_sentinel(const move_sentinel<ConvertibleTo<S>>& s);
-    move_sentinel& operator=(const move_sentinel<ConvertibleTo<S>>& s);
+    explicit constexpr move_sentinel(S s);
+    constexpr move_sentinel(const move_sentinel<ConvertibleTo<S>>& s);
+    constexpr move_sentinel& operator=(const move_sentinel<ConvertibleTo<S>>& s);
 
-    S base() const;
+    constexpr S base() const;
```
ericniebler: I also don't see these changes to `move_sentinel` in p0579. It's a screwup for sure, but we really should file a defect for this.

iterators.tex
CaseyCarter: Yeah, no idea how that happened.

utilities.tex
CaseyCarter: I was hoping to get away with these two; it's clearly in line with the intent of the change.

iterators.tex
CaseyCarter: Yes, these are too big a stretch. I'll pull these out and file an issue.

CaseyCarter:
Repushed without the unauthorized changes to `move_sentinel`'s member functions.

CaseyCarter:
...and without the changes to `tagged`'s constructors that accept `Base`.

Issue #368, opened 2017-03-06
common_iterator's and counted_iterator's const operator* need to be constrained


ericniebler:
Some iterators (e.g., `insert_iterator`) don't have a `const`-qualified `operator*`. When wrapping such an iterator, `counted_iterator` and `common_iterator` also should not have such an overload of `operator*`.

Proposed Resolution
--------------------

1. Change the `common_iterator` class synopsis [common.iterator] as follows:

```diff
 see below operator*();
-see below operator*() const;
-see below operator->() const requires Readable<I>();
+see below operator*() const requires dereferenceable<I const>;
+see below operator->() const requires Readable<I const>();

common_iterator& operator++();
common_iterator operator++(int);
```

2. Change `common_iterator::operator*` [common.iter.op.star] as follows:

```diff
 decltype(auto) operator*();
-decltype(auto) operator*() const;
+decltype(auto) operator*() const requires dereferenceable<I const>;
 1 Requires: !is_sentinel
 2 Effects: Equivalent to: return *iter;
```

3. Change `common_iterator::operator->` [common.iter.op.ref] as follows:

```diff
-see below operator->() const requires Readable<I>();
+see below operator->() const requires Readable<I const>();
 1 Requires: !is_sentinel
 2 Effects: Given an object i of type I
...
```

4. Change the `counted_iterator` class synopsis [counted.iterator] as follows:

```diff
 see below operator*();
-see below operator*() const;
+see below operator*() const requires dereferenceable<I const>;

counted_iterator& operator++();
counted_iterator operator++(int);
```

5. Change `counted_iterator::operator*` [counted.iter.op.star] as follows:

```diff
 decltype(auto) operator*();
-decltype(auto) operator*() const;
+decltype(auto) operator*() const requires dereferenceable<I const>;
 1 Effects: Equivalent to: return *current;
```

ericniebler:
Actually, it can't be `Readable<I const>()` because `common_iterator` and `counted_iterator` can wrap output iterators, which are not `Readable`. It also can't be `Iterator<I const>` because const things are not `WeaklyIncrementable`. Maybe an ad hoc constraint would work here?

ericniebler:
@CaseyCarter for your review.

CaseyCarter:
`Readable<I>()` requires `*` to be valid for a `const I`, so `common_iterator::operator->` is already properly constrained.

[...]

PR #475, opened 2017-07-19
Apply P0740R0 PRs for:

### LINKED ISSUE #366: common_iterator::operator-> is underconstrained
As specified in #318. The "proxy case" constructs an object of type `value_type_t<I>` from `reference_t<I>` but does not require `Constructible<value_type_t<I>, reference_t<I>>()`.

Proposed Resolution
---------------------

Change the class synopsis of `common_iterator` ([common.iterator]) as follows (includes part of the resolution of #368):

```diff
 see below operator*();
 see below operator*() const requires dereferenceable<I const>;
-see below operator->() const requires Readable<I const>();
+see below operator->() const requires see below;
```

To [common.iter.op.ref], add a p3 that reads:

> 3. The expression in the `requires` clause is equivalent to:
> ```c++
> Readable<I const>() &&
>   (requires (const I& i) { i.operator->(); } ||
>    is_reference<reference_t<I>>::value ||
>    Constructible<value_type_t<I>, reference_t<I>>())
> ```

CaseyCarter:
366: common_iterator::operator-> is underconstrained
368: common_iterator’s and counted_iterator’s const operator* need to be constrained

Fixes #366 and #368.

### DIFF
--- a/iterators.tex
+++ b/iterators.tex
@@ -802,7 +802,7 @@ be defined as the iterator's difference type, value type and iterator category, 
     : enable_if<is_object<T>::value, ptrdiff_t> { };
 
   template <class I>
-  struct difference_type<I const> : difference_type<decay_t<I>> { };
+  struct difference_type<const I> : difference_type<decay_t<I>> { };
 
   template <class T>
     requires requires { typename T::difference_type; }
@@ -843,7 +843,7 @@ A \tcode{Readable} type has an associated value type that can be accessed with t
   struct value_type<I> : value_type<decay_t<I>> { };
 
   template <class I>
-  struct value_type<I const> : value_type<decay_t<I>> { };
+  struct value_type<const I> : value_type<decay_t<I>> { };
 
   template <class T>
     requires requires { typename T::value_type; }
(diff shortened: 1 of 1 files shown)


iterators.tex
```diff
@@ -3929,8 +3929,10 @@
     common_iterator& operator=(const common_iterator<ConvertibleTo<I>, ConvertibleTo<S>>& u);
 
     @\seebelow@ operator*();
-    @\seebelow@ operator*() const;
-    @\seebelow@ operator->() const requires Readable<I>;
+    @\seebelow@ operator*() const
```
ericniebler: Didn't LWG want to see `decltype(auto)` in the synopses where appropriate?

CaseyCarter: Yes. I think our guidelines were to use "see below" for long/complicated/metaprogrammed return types, and simply use `decltype(auto)` directly. Fixing.

CaseyCarter:
There's a secondary commit with "see below" cleanups. It's separate for ease of review, but should be squashed on merge.

PR #516, opened 2017-10-26
View interface cmc

CaseyCarter:
A series of changes from a partial review of D0789R1.

### DIFF
--- a/ext/adaptors/ranges-lib.tex
+++ b/ext/adaptors/ranges-lib.tex
@@ -239,7 +239,7 @@ constexpr bool empty() const requires ForwardRange<const D>;
 
 \begin{itemdescr}
 \pnum
-\effects Equivalent to \tcode{return ranges::begin(derived()) == ranges::end(derived());}.
+\effects Equivalent to: \tcode{return ranges::begin(derived()) == ranges::end(derived());}.
 \end{itemdescr}
 
 \indexlibrary{\idxcode{operator bool}!\idxcode{view_interface}}%
@@ -272,7 +272,7 @@ SizedSentinel<sentinel_t<const D>, iterator_t<const D>>;
 
 \begin{itemdescr}
 \pnum
-\effects Equivalent to \tcode{return ranges::end(derived()) - ranges::begin(derived());}.
+\effects Equivalent to: \tcode{return ranges::end(derived()) - ranges::begin(derived());}.
 \end{itemdescr}
 
 \indexlibrary{\idxcode{front}!\idxcode{view_interface}}%
@@ -286,7 +286,7 @@ constexpr decltype(auto) front() const requires ForwardRange<const D>;
 \requires \tcode{!empty()}.
 
 \pnum
-\effects Equivalent to \tcode{return *ranges::begin(derived());}.
+\effects Equivalent to: \tcode{return *ranges::begin(derived());}.
 \end{itemdescr}
 
 \indexlibrary{\idxcode{back}!\idxcode{view_interface}}%
@@ -302,7 +302,7 @@ requires BidirectionalRange<const D> && BoundedRange<const D>;
 \requires \tcode{!empty()}.
 
 \pnum
-\effects Equivalent to \tcode{return *prev(ranges::end(derived()));}.
+\effects Equivalent to: \tcode{return *prev(ranges::end(derived()));}.
 \end{itemdescr}
 
 \indexlibrary{\idxcode{operator[]}!\idxcode{view_interface}}%
@@ -318,7 +318,7 @@ constexpr decltype(auto) operator[](difference_type_t<iterator_t<R>> n) const;
 \requires \tcode{ranges::begin(derived()) + n} is well-formed.
 
 \pnum
-\effects Equivalent to \tcode{return ranges::begin(derived())[n];}.
+\effects Equivalent to: \tcode{return ranges::begin(derived())[n];}.
 \end{itemdescr}
 
 \indexlibrary{\idxcode{at}!\idxcode{view_interface}}%
@@ -333,7 +333,7 @@ constexpr decltype(auto) at(difference_type_t<iterator_t<R>> n) const;
 
 \begin{itemdescr}
 \pnum
-\effects Equivalent to \tcode{return \changed{ranges::begin(derived())}{derived()}[n];}.
+\effects Equivalent to: \tcode{return \changed{ranges::begin(derived())}{derived()}[n];}.
 
 \pnum
 \throws \tcode{out_of_range} if \tcode{n < 0 || n >= ranges::size(derived())}.
@@ -593,7 +593,7 @@ constexpr operator PairLike() const;
 
 \begin{itemdescr}
 \pnum
-\effects Equivalent to \tcode{return PairLike\{begin_, end_\};}.
+\effects Equivalent to: \tcode{return PairLike\{begin_, end_\};}.
 \end{itemdescr}
 
 \rSec4[ranges.subrange.accessors]{\tcode{subrange} accessors}
@@ -605,7 +605,7 @@ constexpr I begin() const;
 
 \begin{itemdescr}
 \pnum
-\effects Equivalent to \tcode{return begin_;}.
+\effects Equivalent to: \tcode{return begin_;}.
 \end{itemdescr}
 
 \indexlibrary{\idxcode{end}!\idxcode{subrange}}%
@@ -615,7 +615,7 @@ constexpr I end() const;
 
 \begin{itemdescr}
 \pnum
-\effects Equivalent to \tcode{return end_;}.
+\effects Equivalent to: \tcode{return end_;}.
 \end{itemdescr}
 
 \indexlibrary{\idxcode{empty}!\idxcode{subrange}}%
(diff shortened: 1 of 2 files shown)


ext/adaptors/ranges-lib.tex
```diff
@@ -14,6 +14,7 @@
 
   // \ref{ranges.view_interface}:
   template <class D>
+    @\added{requires is_class_v<D>}@
```
ericniebler: I don't see where the standard permits `is_class` to be instantiated with an incomplete type.

ext/adaptors/ranges-lib.tex
```diff
@@ -144,7 +145,8 @@
 \begin{codeblock}
 template <class T>
 concept bool ViewableRange =
-  Range<T> && (is_lvalue_reference_v<T> || View<decay_t<T>>); // \seebelow
+  Range<T> && (is_lvalue_reference_v<T> ||
+    @\added{(}@View<decay_t<T>> @\added{\&\& Constructible<decay_t<T>, T>)}@); // \seebelow
```
ericniebler: This is covered by `View`, which requires `Semiregular`.

ext/adaptors/ranges-lib.tex
```diff
@@ -192,7 +195,9 @@
     }
   public:
     constexpr bool empty() const requires ForwardRange<const D>;
-    constexpr explicit operator bool() const requires ForwardRange<const D>;
+    @\added{template <class R = D>}@
+      @\added{requires requires(const R\& r) \{ ranges::empty(r); \}}@
+    constexpr explicit operator bool() const @\removed{requires ForwardRange<const D>}@;
```
ericniebler: I'd rather not.

ext/adaptors/ranges-lib.tex
```diff
@@ -239,17 +246,20 @@
 
 \begin{itemdescr}
 \pnum
-\effects Equivalent to \tcode{return ranges::begin(derived()) == ranges::end(derived());}.
+\effects Equivalent to: \tcode{return ranges::begin(derived()) == ranges::end(derived());}.
```
ericniebler: Perhaps this should be `ranges::empty(derived())`.

ext/adaptors/ranges-lib.tex
```diff
@@ -283,10 +293,10 @@
 
 \begin{itemdescr}
 \pnum
-\requires \tcode{!empty()}.
+\requires \tcode{!\added{ranges::}empty(\added{derived()})}.
```
ericniebler: Unnecessary once we change `empty()`.

ext/adaptors/ranges-lib.tex
```diff
@@ -299,10 +309,10 @@
...
```
ericniebler: Ditto.

ext/adaptors/ranges-lib.tex
```diff
@@ -342,8 +352,8 @@
 \indexlibrary{\idxcode{at}!\idxcode{view_interface}}%
 \begin{itemdecl}
 template <ForwardRange C>
-  requires !View<C> && MoveConstructible<C> &&
-    ConvertibleTo<value_type_t<iterator_t<const D>>, value_type_t<iterator_t<C>>> &&
+  requires !View<C> && @\removed{MoveConstructible<C> \&\&}@
+    ConvertibleTo<@\changed{value_type_t}{reference_t}@<iterator_t<const D>>, value_type_t<iterator_t<C>>> &&
```
ericniebler: Because of guaranteed copy elision?

ext/adaptors/ranges-lib.tex
timsong-cpp: https://timsong-cpp.github.io/cppwp/meta.type.synop#2

ext/adaptors/ranges-lib.tex
CaseyCarter: `View<decay_t<volatile subrange<int*>>>` is satisfied, but `Constructible<subrange<int*>, volatile subrange<int*>>` is not.

ext/adaptors/ranges-lib.tex
CaseyCarter: You prefer to keep it overconstrained? Why?

ext/adaptors/ranges-lib.tex
ericniebler: I really can't be bothered with `volatile`, and I don't want to complicate our interfaces to account for it.

ext/adaptors/ranges-lib.tex
CaseyCarter: That would infinitely recurse if `derived()` does not override `empty`.

ext/adaptors/ranges-lib.tex
CaseyCarter: Yes. (it's in the commit log.)

CaseyCarter: Clarifying: the `MoveConstructible` change is because of guaranteed copy elision, the `value_type_t` to `reference_t` change is to be a bit closer to the `EmplaceConstructible` requirement imposed by the Standard for range constructors of sequence containers.

CaseyCarter: (Doing it properly would require checking allocator construction which is even less likely to be properly constrained.)

ext/adaptors/ranges-lib.tex
CaseyCarter: (This change allows `subrange` to implement `empty` and get `operator bool` from `view_interface`)

ext/adaptors/ranges-lib.tex
CaseyCarter: > I really can't be bothered with volatile, and I don't want to complicate our interfaces to account for it.

I understand that you don't like volatile, but we can't simply pretend that parts of the core language do not exist. Would you prefer blanket wording that forbids instantiating a library template with a type in which `volatile` participates in any way?

CaseyCarter: > we can't simply pretend that parts of the core language do not exist

...despite that LWG has been doing exactly that with `volatile` for years, giving rise to the nasty C++17 NB comment suggesting we should either handle `volatile` properly or forbid it outright.

ericniebler: I prefer to follow LEWG's example and ignore it. When they make a library-wide decision about how to handle it, we can follow suit.

ext/adaptors/ranges-lib.tex
ericniebler: I know. And yes, I have a weak preference to keep it "overconstrained." The fact that some of the members must be templates is an ugly hack that's only necessary because concepts has no support for associated types. I don't want to see that ugly hack spread farther than is necessary.

CaseyCarter: Correction: `template` is only needed here because of GCC 82507. [[temp.inst]/16](http://eel.is/c++draft/temp.inst#16) says I can write:
```c++
constexpr explicit operator bool() const requires requires { ranges::empty(derived()); }
```
and the constraints won't be checked until overload resolution time. Would that be acceptable?

ext/adaptors/ranges-lib.tex
CaseyCarter: Reverted.

ext/adaptors/ranges-lib.tex
ericniebler: That's fine. I have no strong feelings about it now.

ext/adaptors/ranges-lib.tex
```diff
@@ -239,17 +244,19 @@
...
-constexpr explicit operator bool() const requires ForwardRange<const D>;
+constexpr explicit operator bool() const
+  requires @\changed{ForwardRange<const D>}{requires \{ ranges::empty(derived()); \}}@;
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
-\returns \tcode{!empty()}.
+\removed{\returns \tcode{!empty()}.}
+\added{\effects Equivalent to: \tcode{return !ranges::empty(r);}}
```
ericniebler: What is `r`?

ext/adaptors/ranges-lib.tex
```diff
@@ -239,17 +244,19 @@
...
 \pnum
-\effects Equivalent to \tcode{return ranges::begin(derived()) == ranges::end(derived());}.
+\effects Equivalent to: \tcode{return ranges::begin(derived()) == ranges::end(derived());}.
 \end{itemdescr}
 
 \indexlibrary{\idxcode{operator bool}!\idxcode{view_interface}}%
 \begin{itemdecl}
-constexpr explicit operator bool() const requires ForwardRange<const D>;
+constexpr explicit operator bool() const
+  requires @\changed{ForwardRange<const D>}{requires \{ ranges::empty(derived()); \}}@;
```
CaseyCarter: Note that I'm "cheating" the revision marks here a bit: I don't consider whitespace/linebreak changes worth tracking.

ext/adaptors/ranges-lib.tex
CaseyCarter: It's what a pirate says. (cut'n'paste, fixing.)

ext/adaptors/ranges-lib.tex
ericniebler: I don't mark up editorial changes.

CaseyCarter: I thought we were in agreement on this, though I don't recall ever discussing it.

ericniebler: This early in the proceedings, I only want to draw LEWGs attention to substantive changes.

CaseyCarter:
Rebased onto master and cleaned the commit history.

PR #520, opened 2017-11-06
detailed subrange review:

CaseyCarter:
All changes implemented on the subrange_cmc branch of cmcstl2.

* pair-like: ensure that `get<>()` works.
* pair-like-convertible-to:
  * put `!Range<T>` first, because I'm paranoid about requirement recursion.
  * Verify that the result of calling `get` on the actual expression type converts.
* pair-like-convertible-from:
  * Ditto `!Range<T>` first
  * Simplify by removing the `ConvertibleTo` constraints
* iterator-sentinel-pair:
  * Ditto `!Range<T>` first

* Don't name the `subrange_kind` template parameter "Sized" to avoid confusion. Let's use `K`, and replace all of the `B` parameter names with `K`.
* Don't allow `subrange_kind::unsized` when `SizedSentinel<S, I>` holds; it's not needed.
* Fix the last vestiges of the change from `bool` to `subrange_kind`, including using `K == subrange_kind::sized` everywhere instead of `bool(K)`.
* Use perfect forwarding in the `PairLike` constructors now that `pair-like-convertible-to` supports it.
* `next` requires `ForwardIterator`, `prev` requires `BidirectionalIterator`.
* Remove "Requires" element from `subrange(subrange<X, Y, Z> r, difference_type_t<I> n)` constructor specification; it inherits the requirement that `n == distance(r)` from the delegating constructor via "Effects: Equivalent to".
* In `operator PairLike`, use `()` construction as allowed by the `Constructible` requirement.

### DIFF
--- a/ext/adaptors/ranges-lib.tex
+++ b/ext/adaptors/ranges-lib.tex
@@ -470,8 +470,7 @@ namespace std { namespace experimental { namespace ranges { inline namespace v1 
     constexpr bool empty() const;
     constexpr difference_type_t<I> size() const
       requires K == subrange_kind::sized;
-    [[nodiscard]] constexpr subrange next(difference_type_t<I> n = 1) const
-      requires ForwardIterator<I>;
+    [[nodiscard]] constexpr subrange next(difference_type_t<I> n = 1) const;
     [[nodiscard]] constexpr subrange prev(difference_type_t<I> n = 1) const
       requires BidirectionalIterator<I>;
     constexpr subrange& advance(difference_type_t<I> n);
@@ -678,8 +677,7 @@ constexpr difference_type_t<I> size() const
 
 \indexlibrary{\idxcode{next}!\idxcode{subrange}}%
 \begin{itemdecl}
-[[nodiscard]] constexpr subrange next(difference_type_t<I> n = 1) const
-  requires ForwardIterator<I>;
+[[nodiscard]] constexpr subrange next(difference_type_t<I> n = 1) const;
 \end{itemdecl}
 
 \begin{itemdescr}
@@ -690,6 +688,9 @@ auto tmp = *this;
 tmp.advance(n);
 return tmp;
 \end{codeblock}
+
+\pnum \enternote If \tcode{ForwardIterator<I>} is not satisfied, \tcode{next}
+may invalidate \tcode{*this}. \exitnote
 \end{itemdescr}
 
 \indexlibrary{\idxcode{prev}!\idxcode{subrange}}%
--- a/ext/adaptors/ranges-lib.tex
+++ b/ext/adaptors/ranges-lib.tex
@@ -27,7 +27,8 @@ namespace std { namespace experimental { namespace ranges { inline namespace v1 
   @\added{enum class subrange_kind : bool \{ unsized, sized \}; }@
 
   @\added{// \ref{ranges.subrange}:}@
-  @\added{template <Iterator I, Sentinel<I> S = I, subrange_kind Sized = \seebelow>}@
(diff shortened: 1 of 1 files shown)


ext/adaptors/ranges-lib.tex
```diff
@@ -544,40 +559,43 @@
...
+template <ConvertibleTo<I> X, ConvertibleTo<S> Y, subrange_kind Z>
+constexpr subrange(subrange<X, Y, Z> r, difference_type_t<I> n)
+  requires K == subrange_kind::sized;
 \end{itemdecl}
 
 \begin{itemdescr}
 \pnum
-\requires \tcode{bool(B)} is \tcode{false} or \tcode{n == r.size()}.
-
-\pnum
```
ericniebler: Because those requirements are already necessary for `subrange{r.begin(), r.end(), n}`?

ext/adaptors/ranges-lib.tex
```diff
@@ -592,7 +610,7 @@
 \effects Equivalent to:
 \begin{itemize}
 \item If \tcode{StoreSize} is \tcode{true},
-\tcode{subrange\{ranges::begin(r), ranges::end(r), ranges::size(r)\}}.
+\tcode{subrange\{ranges::begin(r), ranges::end(r), distance(r)\}}.
```
ericniebler: Why?

ext/adaptors/ranges-lib.tex
```diff
@@ -625,7 +643,7 @@
 
 \indexlibrary{\idxcode{end}!\idxcode{subrange}}%
 \begin{itemdecl}
-constexpr I end() const;
+constexpr S end() const;
```
ericniebler: Nice catch.

ext/adaptors/ranges-lib.tex
```diff
@@ -659,7 +678,8 @@
 
 \indexlibrary{\idxcode{next}!\idxcode{subrange}}%
 \begin{itemdecl}
-[[nodiscard]] constexpr subrange next(difference_type_t<I> n = 1) const;
+[[nodiscard]] constexpr subrange next(difference_type_t<I> n = 1) const
+  requires ForwardIterator<I>;
```
ericniebler: Why?

ext/adaptors/ranges-lib.tex
CaseyCarter: Yes, "Effects equivalent to" inherits requirements ([[structure.specifications]/4](http://eel.is/c++draft/structure.specifications#4)).

ext/adaptors/ranges-lib.tex
CaseyCarter: So the implementation isn't required to produce nasty conversion warnings when `ranges::size` returns an unsigned type.

ext/adaptors/ranges-lib.tex
CaseyCarter: To avoid spooky action at a distance. This copies the iterator and increments the copy; if `I` is single-pass, calling `next` will invalidate `*this`. Other options are (1) have `next` be non-`const` when `I` is single-pass, or (2) I just get over the spooky action at a distance and we add a note telling users that calling `next` invalidates single-pass subranges.

ericniebler: My vote goes for (2). It's true generally that operations on input ranges are not equality-preserving.

CaseyCarter: There's a bit of a safety argument that users should call `advance` instead of `next` with input ranges to avoid having invalidated ranges lying around. I'll keep it in mind as we gather usage experience for `subrange`, but revert this for now.

PR #522, opened 2017-11-06
range adaptor types do not have nested type aliases for iterator or const_iterator

ericniebler:


ext/adaptors/ranges-lib.tex
```diff
@@ -3186,46 +3208,72 @@
...
-respectively.
+\effects Equivalent to\added{:}
+\begin{codeblock}
+@\added{return} \changed{sentinel}{\xname{sentinel}<\textit{simple-view}<R>{>}}\{*this\}\added{;}@
+\end{codeblock}
+  and
+\begin{codeblock}
+@\added{return} \changed{const_sentinel}{\xname{sentinel}<true>}\{*this\}\added{;}@
+\end{codeblock}
+for the first and second overload, respectively.
```
CaseyCarter: I'm wary of having two members that do the same thing in the `simple-view<R>` case; it increases code bloat if the linker doesn't perform identical comdat folding, and just causes confusion in general. I would prefer that the non-`const` overload requires `!simple-view<R>` and just disappears. (This is admittedly a persnickety detail this early in the design process.)

CaseyCarter (approved):
All good, except the missing const.

PR #527, opened 2017-12-27
minor view_interface tweaks

CaseyCarter:
* add `is_class_v` constraint
* `s/The view_interface/The class template view_interface/`
* Don't use a trailing-requires-clause for template `data`

The `is_class_v` constraint borders on a design change, but I think we can call it a wording fix to realize the intended design with a straight face and skip LEWG.

### DIFF
--- a/ext/adaptors/ranges-lib.tex
+++ b/ext/adaptors/ranges-lib.tex
@@ -14,6 +14,7 @@ namespace std { namespace experimental { namespace ranges { inline namespace v1 
 
   // \ref{ranges.view_interface}:
   template <class D>
+    @\added{requires is_class_v<D>}@
   class view_interface;
 
   @\removed{// 10.7.2.1:}@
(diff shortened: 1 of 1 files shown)


ext/adaptors/ranges-lib.tex
```diff
@@ -212,7 +214,8 @@
     @\removed{constexpr bool operator!() const requires ForwardRange<const D>;}@
 
     @\added{template <RandomAccessRange R = const D>}@
-      @\added{constexpr auto data() const requires is_pointer_v<iterator_t<R>>;}@
+        @\added{requires is_pointer_v<iterator_t<R>{>}}@
+      @\added{constexpr auto data() const;}@
```
ericniebler: I'm just curious, why?

CaseyCarter: Simply for consistency: we have not used trailing requires clauses on function templates before.

PR #540, opened 2018-02-09
Strike [std2.numerics]

CaseyCarter:
P0898 merges this concept into namespace std and renames it `UniformRandomBitGenerator` to agree with the renaming of the requirements set in C++17.


ext/merge2std2/algorithms.tex
```diff
@@ -542,13 +542,13 @@
   // \ref{std2.alg.random.shuffle}, shuffle:
   template <RandomAccessIterator I, Sentinel<I> S, class Gen>
     requires Permutable<I> &&
-      UniformRandomNumberGenerator<remove_reference_t<Gen>> &&
+      UniformRandom@\oldtxt{Number}\newtxt{Bit}@Generator<remove_reference_t<Gen>> &&
```
ericniebler: I've been using `\added` and `\removed` (or in this case, `\changed{Number}{Bit}`) for edits to Ranges TS text.

CaseyCarter: Fixed.

CaseyCarter: Sorry for the delay, I just realized my Chrome notifications have been broken all day.

PR #541, opened 2018-02-09
Fix iterator adaptor conversions:

CaseyCarter:
Don't use "abbreviated function templates" since they are not in C++20.

### DIFF
--- a/ext/merge2std2/iterators.tex
+++ b/ext/merge2std2/iterators.tex
@@ -2223,8 +2223,10 @@ namespace @\changed{std \{ namespace experimental \{ namespace ranges}{std2}@ { 
 
     constexpr reverse_iterator();
     explicit constexpr reverse_iterator(I x);
-    constexpr reverse_iterator(const reverse_iterator<ConvertibleTo<I>>& i);
-    constexpr reverse_iterator& operator=(const reverse_iterator<ConvertibleTo<I>>& i);
+    @\added{template <ConvertibleTo<I> U>}@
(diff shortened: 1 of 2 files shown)


ext/merge2std2/iterators.tex
```diff
@@ -2223,8 +2223,10 @@
 
     constexpr reverse_iterator();
     explicit constexpr reverse_iterator(I x);
-    constexpr reverse_iterator(const reverse_iterator<ConvertibleTo<I>>& i);
-    constexpr reverse_iterator& operator=(const reverse_iterator<ConvertibleTo<I>>& i);
+    @\added{template <ConvertibleTo<I> U>}@
+      constexpr reverse_iterator(const reverse_iterator<\changed{ConvertibleTo<I>}{U}>& i);
```
CaseyCarter: Missing `@`s here.

ext/merge2std2/ranges-lib.tex
```diff
@@ -22,7 +22,7 @@
 \rSec1[std2.ranges.decaycopy]{decay_copy}
 
 \ednote{TODO: Replace the definition of [thread.decaycopy] with this definition, and
-move it where both standard libraries can make use of it. Location TBD.}
+move it where both standard libraries can make use of it. Location TBD.}
```
CaseyCarter: I don't know what that old character was, but latex turned it into mojibake.

ericniebler:
Thanks. I caught all the `foo<auto>` cases, but not these.

PR #545, opened 2018-02-10
Remove deprecated rvalue behavior in range.access

CaseyCarter:


ericniebler:
If I am correctly recalling our discussion about this, we (or I) wanted to permit passing an rvalue to `std2::begin` so long as ADL finds a `begin` overload that works with rvalues. We needed to change the poison pill to get this behavior, but it gives us a back-door way to detect that a particular range type has iterators that can outlive the range.

This PR removes that possibility by preventing the passing of rvalues to `std2::begin`. I would rather wait to make any changes until we have a final design.

CaseyCarter:
I would prefer to remove the behavior that's broken *now*, and add in behavior that's not broken when that design is complete. This PR doesn't preclude the possibility of future changes, but it does keep the design space clear. Even if the result of this PR makes it into a published standard, future changes would be possible since they would turn ill-formed code into good code.

I'd also like to remove any chance that the current deprecated behavior makes its way into the WD even if we're both hit by the same bus.

ericniebler:
I'm opting to not take this change for the pre-meeting mailing. I don't want to change something that we know we're going to change again. I will add an editorial note explaining our intentions wrt to the deprecated overloads, and add an issue to make sure this doesn't get lost if we're on the same ill-fated bus.

CaseyCarter:
No problem - I honestly couldn't remember the result of our earlier discussion when I wrote up this PR. There's clearly No Consensus For Change At This Time.

PR #557, opened 2018-05-03
[P0896] indirect_result fixes

CaseyCarter:
* variadic instead of single-function-type parameter
* require the `Is...` to be `Readable` before using `reference_t` on them

### DIFF
--- a/ext/merge2std2/iterators.tex
+++ b/ext/merge2std2/iterators.tex
@@ -1034,11 +1034,11 @@ callable objects~(\cxxref{func.def}) as arguments.
     StrictWeakOrder<F&, reference_t<I1>, reference_t<I2>> &&
     StrictWeakOrder<F&, iter_common_reference_t<I1>, iter_common_reference_t<I2>>;
 
-  template <class> struct indirect_result@\oldtxt{_of}@ { };
+  @\oldtxt{template <class> struct indirect_result_of \{ \};}@
 
   template <class F, class... Is>
-    requires Invocable<F, reference_t<Is>...>
-  struct indirect_result@\oldtxt{_of}@<F(Is...)> :
+    requires @\newtxt{(Readable<Is> \&\& ...) \&\&}@ Invocable<F, reference_t<Is>...>
+  struct indirect_result@\oldtxt{_of}@<F@\oldtxt{(}\newtxt{, }@Is...@\oldtxt{)}@> :
     @\oldtxt{result_of<F(reference_t<Is>\&\&...)>}\newtxt{invoke_result<F, reference_t<Is>...>}@ { };
 \end{codeblock}
 
--- a/ext/merge2std2/ranges-lib.tex
+++ b/ext/merge2std2/ranges-lib.tex
@@ -148,15 +148,16 @@ namespace std { @\oldtxt{namespace experimental \{}@
     template <class F, class I1, class I2 = I1>
     concept @\oldtxt{bool}@ IndirectStrictWeakOrder = @\seebelow@;
 
-    template <class> struct indirect_result@\oldtxt{_of}@;
+    template <class@\newtxt{, class...}@>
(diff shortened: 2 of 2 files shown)


ext/merge2std2/iterators.tex
```diff
@@ -1034,11 +1034,12 @@
...
 
-  template <class> struct indirect_result@\oldtxt{_of}@ { };
+  template <class@\newtxt{, class...}@>
+  struct indirect_result@\oldtxt{_of}@ { };
 
   template <class F, class... Is>
-    requires Invocable<F, reference_t<Is>...>
-  struct indirect_result@\oldtxt{_of}@<F(Is...)> :
+    requires @\newtxt{(Readable<Is> \&\& ...) \&\&}@ Invocable<F, reference_t<Is>...>
+  struct indirect_result@\oldtxt{_of<F(Is...)>}@ :
```
ericniebler: Nope, should be: `struct indirect_result@\oldtxt{_of<F(Is...)>}\newtxt{<F, Is...>}@`. Below, also.

ext/merge2std2/ranges-lib.tex
```diff
@@ -148,15 +148,16 @@
...
-      requires Invocable<F, reference_t<Is>...>
-    struct indirect_result@\oldtxt{_of}@<F(Is...)>;
+      requires @\newtxt{(Readable<Is> \&\& ...) \&\&}@ Invocable<F, reference_t<Is>...>
+    struct indirect_result@\oldtxt{_of<F(Is...)>}@;
 
-    template <class F>
+    template <class F@\newtxt{, class... Is}@>
     using indirect_result@\oldtxt{_of}@_t
-      = typename indirect_result@\oldtxt{_of}@<F>::type;
+      = typename indirect_result@\oldtxt{_of}@<F, Is...>::type;
```
CaseyCarter: Markup?

ext/merge2std2/ranges-lib.tex
```diff
@@ -148,15 +148,16 @@
     template <class F, class I1, class I2 = I1>
     concept @\oldtxt{bool}@ IndirectStrictWeakOrder = @\seebelow@;
 
-    template <class> struct indirect_result@\oldtxt{_of}@;
+    template <class@\newtxt{, class...}@>
+    struct indirect_result@\oldtxt{_of}@ { };
```
CaseyCarter: Markup?

ext/merge2std2/iterators.tex
```diff
@@ -1034,11 +1034,12 @@
     StrictWeakOrder<F&, reference_t<I1>, reference_t<I2>> &&
     StrictWeakOrder<F&, iter_common_reference_t<I1>, iter_common_reference_t<I2>>;
 
-  template <class> struct indirect_result@\oldtxt{_of}@ { };
+  template <class@\newtxt{, class...}@>
+  struct indirect_result@\oldtxt{_of}@ { };
```
CaseyCarter: Markup?

ext/merge2std2/ranges-lib.tex
CaseyCarter: Better yet, just remove the definition. It doesn't need to be both here *and* in the detailed specification.

ericniebler: ... or just put the definition in the synopsis. After all, the definition is just `{ }`.

CaseyCarter: That's probably how I would have written this in a green field, but with no real reason to prefer one style or the other I went with what was in R0: non-defining declaration here, definition in the detailed spec.

ericniebler: No worries, I'll fix it in P1037.

CaseyCarter: Fixed.

PR #558, opened 2018-05-04
[P0898] Finish making advance et al. "implicit CPOs"

CaseyCarter:
* merge `distance(SizedRange)` into `distance(Range)`


ext/merge2std2/iterators.tex
```diff
@@ -1648,24 +1648,30 @@
 
 \begin{itemdecl}
 template <Range R>
-  constexpr difference_type_t<iterator_t<R>> @\oldtxt{operator()}\newtxt{distance}@(R&& r) const;
+  constexpr difference_type_t<iterator_t<R>> @\oldtxt{operator()}\newtxt{distance}@(R&& r) @\oldtxt{const}@;
 \end{itemdecl}
 
 \begin{itemdescr}
 \effects
```
CaseyCarter: These paragraphs could use some `\pnum`s

PR #559, opened 2018-05-05
[1037] review updates:

CaseyCarter:
* Base on proper rev of P0896 and P0944
* There is no `ranges-lib.tex`
* Update bib entry for P0898
* I fixed `indirect_result` in P896R1
* P896R1 merged the `SizedRange` overload of `distance` into the `Range` overload
* `pointer` can't be `void` for iterators that *do* support `operator->`
* Strip trailing `const` from `advance` et al
* Add .gitignore


ext/integrate/D1037.tex
```diff
@@ -97,7 +97,7 @@
 % \input{lib-intro}
 % \input{utilities}
 \input{iterators}
-\input{ranges-lib}
+% \input{ranges-lib}
```
ericniebler: That was an oversight. I just added my ranges-lib.tex. Please rebase and revert this change.

ext/integrate/iterators.tex
```diff
@@ -150,14 +150,14 @@
...
-  template <class, class... Is> struct indirect_result@\added{ \{ \}}@;
+  template <class, class...> struct indirect_result { };
 
   template <class F, class... Is>
     requires (Readable<Is> &&...) && Invocable<F, @\added{iter_}@reference_t<Is>...>
   struct indirect_result<F, Is...>;
 
   template <class F, class... Is>
-  using indirect_result_t = typename indirect_result<F>::type;
+  using indirect_result_t = typename indirect_result<F, Is...>::type;
```
ericniebler: Nice catch.

